PR #2, opened 2021-08-06
Reorder material for 1st day, add functions exercise

hageboeck:
- As discussed, the functions part was moved right after pointers and references.
- There's a small exercise for pass by copy / pass by reference
- Some small tweaks / fix typos.
- Added code highlighting to slides where symbolic stacks are shown.

Note:
On the slide about arrays and pointers, there was an array cookie/pointer after the array on the symbolic stack. Since this isn't a dynamic allocation, there shouldn't be anything but padding bytes, so I removed it. It won't make a difference for the lecture.


hageboeck:
Close in  favour of #3

PR #4, opened 2021-08-10
Few improvements for OO part

hageboeck:
- Fix a few typos, reword a few sentences
- Tweak layout and alignments on a few slides
- Use `float` literals instead of `int` to show how it's done
- **Please check this:** Add an `override` on a c++98 slide to give a good example of modern style. Should I update the c++98 label in the top corner?

@sponce

### DIFF
--- a/talk/objectorientation.tex
+++ b/talk/objectorientation.tex
@@ -439,7 +439,7 @@
     \begin{itemize}
     \item special functions building/destroying an object
     \item a class can have several constructors
-    \item the constructors have the name of the class
+    \item the constructors have the name as the class
     \item same for the destructor with a leading $\sim$
     \end{itemize}
   \end{block}
@@ -565,7 +565,7 @@
   \frametitleii{Default Constructor}
   \begin{block}{Idea}
     \begin{itemize}
-    \item avoid writing explicitely default constructors
+    \item avoid writing explicitly default constructors
     \item by declaring them as default
     \end{itemize}
   \end{block}
@@ -573,7 +573,7 @@
     \begin{itemize}
     \item when no user defined constructor, a default is provided
     \item any user defined constructor disables default ones
-    \item but they can be enforced.
+    \item but they can be enforced
     \item rule can be more subtle depending on members
     \end{itemize}
   \end{block}
@@ -650,7 +650,7 @@
 \end{frame}
 
 \begin{frame}[fragile]
-  \frametitleii{Calling contructors}
+  \frametitleii{Calling constructors}
   \begin{block}{After object declaration, arguments within \{\}}
     \begin{multicols}{2}
       \begin{cppcode*}{gobble=4}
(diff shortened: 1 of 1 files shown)


talk/objectorientation.tex
```diff
@@ -132,7 +132,7 @@
     \begin{itemize}
     \item overloading is authorized and welcome
     \item signature is part of the method identity
-    \item but not the return code
+    \item but not the return type
```
sponce, line 135: No I really mean type here. This should not be changed.

sponce:
```suggestion
    \item but not the return code
```

talk/objectorientation.tex
```diff
@@ -634,7 +634,8 @@
     \end{itemize}
   \end{block}
   \begin{exampleblock}{Practically}
-    \begin{cppcode}
+   \begin{minipage}{0.7\textwidth}
+    \begin{cppcode*}{highlightlines={2,7}}
```
sponce, line 638: Does this compile ? Where is the \end{minipage} statement ?

sponce:
```suggestion
    \begin{cppcode}
```

talk/objectorientation.tex
```diff
@@ -981,12 +982,13 @@
   \frametitlegb{Virtual methods}
   \begin{block}{the concept}
     \begin{itemize}
-    \item methods can be declared {\it virtual}
-    \item for these, the most precise object is always considered
+    \item methods can be declared \mintinline{cpp}{virtual}
+    \item for these, the most precise (derived) object is always considered
```
sponce, line 986: Let's fit on one line

sponce:
```suggestion
    \item for these, the most derived object is always considered
```

talk/objectorientation.tex
hageboeck: > No I really mean >>type<< here. This should not be changed.

Was this a typo in the above message? So we use code like "exit code".

talk/objectorientation.tex
hageboeck: > Does this compile ? Where is the \end{minipage} statement ?

It's after the code block, but it was only needed to limit the width of the code block:
https://github.com/hageboeck/cpluspluscourse/blob/8068490708f516569609ab8f7f7d9408e89a7fb7/talk/objectorientation.tex#L650

If no lines are highlighted, you don't need the minipage.

I think I'll take it out.

talk/objectorientation.tex
sponce: Yes typo... now fixed It's really the type that is important here. I'm not sure actually what you meant by "code"

talk/objectorientation.tex
hageboeck: Ouch, it's in another commit. I'll fix it.

talk/objectorientation.tex
hageboeck: OK, agree on using "return type".

hageboeck:
Ok, the things that were discussed above are done. Ready to merge.

PR #15, opened 2021-08-31
some changes during Aug2021 course

roiser:
I open a PR with some changes we discover during the Aug2021 course lecturing

(sorry my editor removed some trailing spaces, so the first commit is actually more than it really is...)


roiser:
- comment: use function default parameters on the  first day

hageboeck:
If @roiser doesn't want to add more, I suggest to merge #16 instead, as it contains the two commits from this PR.
Or you merge this and I rebase.

PR #23, opened 2021-08-31
Avoid creating shared library in code/hello

eguiraud:
Several students were confused by that extra file and/or forgot to
`export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:.` and/or asked what that
magic spell did.

Explanation of the compilation chain is on the last day, we could
leave creation of shared objects and LD_LIBRARY_PATH for that day.

If people agree with this change I can apply similar patches to all exercises of days 1, 2 and 3.

### DIFF
--- a/code/hello/Makefile
+++ b/code/hello/Makefile
@@ -1,11 +1,8 @@
-all: libhello.so hello
+all: hello
 solution: all
 
 clean:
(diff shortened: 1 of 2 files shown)


sponce:
This one I would close as the extra complication of having this library was precisely meant to check whether '.' was in LD_LIBRARY_PATH as we need it later.

PR #24, opened 2021-08-31
add a section on static assert, improve SFINAE with void_t

theanalyst:
Use void_t for sfinae example so that we don't fixate on return types. Another option is just adding a `void()` in the end of decltype expr (eg: https://gcc.godbolt.org/z/9qGoe173x)

 Also introduce static_assert, I added this in the constexpr section though could make sense only in the expert section. 
Since the material is mostly template meta programming and probably not too relevant feel free to suggest/reject/modify this ;)

### DIFF
--- a/talk/expert.tex
+++ b/talk/expert.tex
@@ -383,7 +383,7 @@
 \end{frame}
 
 \begin{frame}[fragile]
-  \frametitleii{Using SFINAE for instrospection}
+  \frametitleii{Using SFINAE for introspection}
   \begin{block}{The main idea}
     \begin{itemize}
     \item use a template specialization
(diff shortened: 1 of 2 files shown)


talk/expert.tex
```diff
@@ -403,11 +403,45 @@
...
+      struct A{}; struct B{void foo();}; struct C{int foo();};
+      static_assert(!hasFoo<A>::value, "A has no Foo()");
+      static_assert(hasFoo<B>::value, "B has Foo()");
+      // static_assert(hasFoo<C>::value, "C has Foo()"); fails assertion!
     \end{cppcode*}
   \end{exampleblock}  
 \end{frame}
 
+\begin{frame}[fragile]
+  \frametitleii{Using \texttt{void\_t} to simplify SFINAE expressions \hfill \cpp17}
```
sponce, line 415: frametitleii will put C++11 in the title but you add also C++ 17 manually. This is contradictory.
I think you want to use frametitleia (sorry for the weird leet coding but latex does not allow numbers in commands and frametitleseventeen was too long for me...)

talk/expert.tex
```diff
@@ -403,11 +403,45 @@
       struct hasFoo<T, decltype(std::declval<T>().foo())>
         : std::true_type {};
 
-      std::cout << hasFoo<MyType>::value << std::endl;
+      struct A{}; struct B{void foo();}; struct C{int foo();};
+      static_assert(!hasFoo<A>::value, "A has no Foo()");
+      static_assert(hasFoo<B>::value, "B has Foo()");
+      // static_assert(hasFoo<C>::value, "C has Foo()"); fails assertion!
```
sponce, line 409: Nice but does not fit at all in the slide ! Maybe just add a `\small` before the code section

talk/expert.tex
```diff
@@ -403,11 +403,45 @@
...
+      struct hasFoo<T, std::void_t<decltype(std::declval<T>().foo())>>
+      : std::true_type {};
+
+      struct A{}; struct B{void foo()}; struct C{int foo();};
+      static_assert(!hasFoo<A>::value, "A has no Foo()");
+      static_assert(hasFoo<B>::value, "B has Foo()");
+      static_assert(hasFoo<C>::value, "C has Foo()");
+    \end{cppcode*}
+  \end{exampleblock}
+\end{frame}
```
sponce, line 443: Again, does not fit in the slide. And by far this time, things are cut after line 434.
Now first 2 lines can be put on 1, last bullet could be changed to "Used to ..." to be on one line and then you can put hte code in `\small` font. Not sure it will be enough though

talk/morelanguage.tex
```diff
@@ -118,6 +118,29 @@
...
+  \pause
+  \begin{exampleblock}{Example}
+    \begin{cppcode*}{}
+      constexpr int f(int x) {
+        return x > 1 ? x * f(x - 1) : 1;
+      }
+      static_assert(f(5)==120,"Expected f(5) to be 120!");
+    \end{cppcode*}
+  \end{exampleblock}
+\end{frame}
```
sponce, line 141: It does fit this time, but barely. Easy to fix this time, by having the second bullet on a single line : "The expression has to be a constexpr boolean expression". I would also probably put the second part of the first point as a subbullet, gaining both a bit of space and readibility

talk/morelanguage.tex
```diff
@@ -177,8 +201,8 @@
     constexpr DimLength km(1, 'k');
     constexpr float km_y = km.get('y');
     constexpr float km_i = km.get('i');
-    std::cout << "1 km = " << km_y << " yards\n"
-              << "     = " << km_i << " inches\n";
+    // static_assert(km_y == 1000, "Expected a km to be 1000 yards"); Fails compilation
+    static_assert(km_y == 1093, "Expected a km to be 1093 yards! ");
```
sponce, line 205: These 2 lines are too long and do not fit in the slide. Try switching to `\small` once more

sponce (changes requested):

PR #25, opened 2021-09-01
Add codespell pre commit hook and some others

klieret:
* Fixed some typos 
* Added corresponding pre-commit hook
* Stripped trailing whitespace
* Added corresponding pre-commit hook

Depending on whether we want all of that, we can also take a subset of these commits. But I'm a big fan of pre-commit in general. I could also add it as a CI here so that it gets checked on every push/PR.


klieret:
I started over: https://github.com/hsf-training/cpluspluscourse/pull/26 (didn't want to handle merge conflicts due to old version)

PR #26, opened 2021-09-01
Add codespell pre commit hook 2

klieret:
(started new because I had based the previous branch on an outdated version and the merge conflicts were annoying)

    Fixed some typos
    Added corresponding pre-commit hook
    Stripped trailing whitespace
    Added corresponding pre-commit hook

Depending on whether we want all of that, we can also take a subset of these commits. But I'm a big fan of pre-commit in general. I could also add it as a CI here so that it gets checked on every push/PR.


sponce:
That's great. I do not dare merging it right away as we are in the middle of the course and I do not want to risk breaking stuff. But I'll definitely merge it at the end of the week !

PR #28, opened 2021-09-01
Added clean target and solution to asan example

sponce:


sponce:
@hageboeck can you check that the solution is what you expect. In particular for cases 2 and 3, that's how I would write it for this simple example, but maybe you were expecting a usage of move semantic returning a unique_ptr ?

hageboeck (approved):
It's good. It's not what I expected (I didn't think about making the array longer, but stopping the loop earlier), but that's fine!

Maybe we use `std::size(a)` in the loop?

sponce:
I've finally stopped the loop earlier, as it gives 2 opportunities : showing people `std::size` as you mention and also mentioning again that C strings are 0 terminated. In other words, you have to stop the loop at `std::size(a)-1` or your erase the last 0 and print garbage !

PR #30, opened 2021-09-02
Add two slides about clang-tidy

bernhardmgruber:
This adds two slides about clang-tidy. It discusses the `compile_commands.json` and the `.clang-tidy` file. It shows how to run `clang-tidy` and the `run-clang-tidy` wrapper scripts. It includes an exercise where the students should generate a `compile_commands.json` file for any exercise and run clang-tidy on it, inspet the output and apply automatic fixes. Then they should revert the changes again.

PR #34, opened 2021-09-02
Add two slides about clang-format

bernhardmgruber:
This PR adds two slides about clang-format. It explains how to create a `.clang-format` file, how to run clang-format on a single file with a default style and based on a `.clang-format` file. The PR includes an exercise where the students should format one of the examples with a default style, then create a `.clang-format` file and format multiple files.

### DIFF
--- a/talk/tools.tex
+++ b/talk/tools.tex
@@ -81,6 +81,47 @@
   \end{minted}
 \end{frame}
 
+\subsection[format]{Code formatting}
+
+\begin{frame}[fragile]
+\frametitle{clang-format}
+\begin{block}{.clang-format}
+	\begin{itemize}
+		\item file describing your formatting preferences
+		\item should be checked-in at the repository root (project wide)
+		\item \mintinline{bash}{clang-format -style=LLVM -dump-config >} \\
+		  \mintinline{bash}{.clang-format}
+		\item adapt style options with help from: \url{https://clang.llvm.org/docs/ClangFormatStyleOptions.html}
+	\end{itemize}
+\end{block}
+\begin{block}{Run clang-format}
+	\begin{itemize}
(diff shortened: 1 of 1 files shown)


sponce:
Just a tiny comment : I would move these 2 slides just after code management. I find it more logical to start with everything about textual content before going to compilation, debugging, etc... Otherwise very nice. Thanks a lot

bernhardmgruber:
> I would move these 2 slides just after code management.

Done!

I furthermore put all the bash statements into `\mintinline`, as suggested by @hageboeck in #30.

talk/tools.tex
```diff
@@ -81,6 +81,47 @@
...
+\end{frame}
+
+\begin{frame}[fragile]
+\frametitle{clang-format}
+\begin{alertblock}{Exercise Time}
+	\begin{itemize}
+		\item go to any example
+		\item format code with: \mintinline{bash}{clang-format --style=GNU -i <file.cpp>}
+		\item inspect changes, try \mintinline{bash}{git diff}
+		\item revert changes using \mintinline{bash}{git checkout <file.cpp>}
```
hageboeck, line 115:
```suggestion
		\item revert changes using \mintinline{bash}{git checkout -- <file.cpp>}
```
?
In my git, I need the `--` to really check out a file. It might otherwise conflict with branch names.

bernhardmgruber: I just tried that and both work. I can do:
`git checkout tools.tex`
`git checkout -- tools.tex`
Both seem to work for me. But I indeed wondered then what would happen if I had a branch named `tools.tex`, so I tried it, and your solution keeps working. I will change the slides.

bernhardmgruber: Done.

PR #36, opened 2021-09-03
friday  changes

roiser:


talk/expert.tex
```diff
@@ -240,7 +240,7 @@
     template<typename T>
     T&& forward(typename std::remove_reference<T>::type& t)
       noexcept {
-      return static_cast<T&&>(t);
+      return std::move(t);
```
sponce, line 243: One would need `std::move<T>(t)` here, as we need to force the reference collapsing when needed. So I wonder now whether it's not better to leave it with static_cast.

On top, this seems to be the only change remaining in this MR after the fixes by @hageboeck, so maybe I just close it ?

sponce: @roiser ping. Can I close ?

roiser: yes, pls close

PR #59, opened 2021-11-16
add control statements example

roiser:
A little example on control  statements, this may serve as the first example. It's rather easy and should give students confidence.

### DIFF
--- a/code/control/README.md
+++ b/code/control/README.md
@@ -3,6 +3,6 @@
 * Replace the C++98-style for loop with a range-based loop
 * Replace the if / else statement with a conditional operator
 * Replace the if / else statement with a switch statement
-  * Note: You will need to change the signature of the "isodd" function to make this work. Have a look at the return type of the function. 
+  * Note: You will need to change the signature of the "isodd" function to make this work. Have a look at the return type of the function.
 * Replace the for loop with a while loop - does it make sense?
 * Replace the for loop with a do / while loop - does it make sense?
--- a/code/control/README.md
+++ b/code/control/README.md
@@ -3,5 +3,6 @@
 * Replace the C++98-style for loop with a range-based loop
 * Replace the if / else statement with a conditional operator
 * Replace the if / else statement with a switch statement
+  * Note: You will need to change the signature of the "isodd" function to make this work. Have a look at the return type of the function. 
 * Replace the for loop with a while loop - does it make sense?
 * Replace the for loop with a do / while loop - does it make sense?
--- a/code/control/README.md
+++ b/code/control/README.md
@@ -1,6 +1,6 @@
 ## Instructions
 
-* Replace the C++98 style for loop with a range based loop
+* Replace the C++98-style for loop with a range-based loop
 * Replace the if / else statement with a conditional operator
 * Replace the if / else statement with a switch statement
 * Replace the for loop with a while loop - does it make sense?
--- a/code/control/README.md
+++ b/code/control/README.md
@@ -1,0 +1,7 @@
+## Instructions
+
+* Replace the C++98 style for loop with a range based loop
+* Replace the if / else statement with a conditional operator
+* Replace the if / else statement with a switch statement
+* Replace the for loop with a while loop - does it make sense?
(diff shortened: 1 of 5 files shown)


code/control/control.cpp
```diff
@@ -0,0 +1,21 @@
+#include <cstdio>
+
+int isodd(int i) { return i % 2; }
```
hageboeck, line 3:
```suggestion
bool isodd(int i) { return i % 2; }
```

code/control/solution/control.sol.cpp
```diff
@@ -0,0 +1,76 @@
+#include <cstdio>
+
+int isodd(int i) { return i % 2; }
```
hageboeck, line 3:
```suggestion
bool isodd(int i) { return i % 2; }
```

code/control/solution/control.sol.cpp
```diff
@@ -0,0 +1,76 @@
...
+    int num = numbers[i];
+    if (isodd(num)) {
+      sum_odd += num;
+    } else {
+      sum_eve += num;
+    }
+  }
+  printf("Sums: odd = %d, even = %d\n", sum_odd, sum_eve);
+
+  // range based loop
```
hageboeck, line 22:
```suggestion
  // range-based loop
```

code/control/README.md
```diff
@@ -0,0 +1,7 @@
+## Instructions
+
+* Replace the C++98 style for loop with a range based loop
```
hageboeck, line 3:
```suggestion
* Replace the C++98-style for loop with a range-based loop
```

code/control/solution/control.sol.cpp
```diff
@@ -0,0 +1,76 @@
...
+  for (int i : numbers) {
+    isodd(i) ? sum_odd += i : sum_eve += i;
+  }
+  printf("Sums: odd = %d, even = %d\n", sum_odd, sum_eve);
+
+  // switch statement
+  sum_odd = 0;
+  sum_eve = 0;
+  for (int i : numbers) {
+    switch (isodd(i)) {
```
hageboeck, line 46:
```suggestion
    switch ( int(isodd(i)) ) {
```
Now I see why it was `int` ... 🙂

code/control/solution/control.sol.cpp
```diff
@@ -0,0 +1,76 @@
...
+  }
+  printf("Sums: odd = %d, even = %d\n", sum_odd, sum_eve);
+
+  // while loop
+  sum_odd = 0;
+  sum_eve = 0;
+  int i = 0;
+  while (i < nnums) {
+    int num = numbers[i++];
+    isodd(num) ? sum_odd += num : sum_eve += num;
```
hageboeck, line 63: Maybe we can try one-statement if bodies? It looks cleaner to me, but it's a matter of taste ...:
```suggestion
    if (isodd(num)) sum_odd += num;
    else sum_eve += num;
```

hageboeck (commented):
Collection of random comments. Feel free to discard some/all.

code/control/control.cpp
roiser: As you found out below the reason why it is int is that the move to the switch statement won't work. But, as we privately discussed, maybe a good solution is to have it bool here and in the description when asking the students to move to the switch statement tell them that the "isodd" function won't work anymore and they would need to slightly change it.

code/control/solution/control.sol.cpp
roiser: I guess here we need to leave the int return type then.

code/control/solution/control.sol.cpp
roiser: Actually I would leave the signature above then.

code/control/solution/control.sol.cpp
roiser: Sebastien is using this syntax in his slides so wanted to repeat it here in the example code.

sponce:
This is now superseded by #76 where I extended this branch with better cases for switch and while. @roiser if you can review it and confirm it works fine

PR #60, opened 2021-11-16
Mention scoped_lock instead of lock_guard.

### LINKED ISSUE #45: present scoped_lock instead of lock_guard
Or keep both and explain the differences

hageboeck:
Fix #45.

- Change `lock_guard` mentions to `scoped_lock`.
- Add a slide with best practices (== prefer `scoped_lock`)

### DIFF
--- a/talk/concurrency.tex
+++ b/talk/concurrency.tex
@@ -213,18 +213,18 @@
 \end{frame}
 
 \begin{frame}[fragile]
-  \frametitleii{Mutexes}
+  \frametitleii{Mutexes and Locks}
   \begin{block}{Concept}
     \begin{itemize}
-    \item a lock to serialize access to a non-atomic piece of code
+    \item Use locks to serialize access to a non-atomic piece of code
     \end{itemize}
   \end{block}
   \pause
   \begin{block}{The objects}
     \begin{description}[labelwidth=1.8cm]
     \item[std::mutex] in the mutex header
-    \item[std::lock\_guard] for an RAII version of it
-    \item[std::unique\_lock] same and can be released/relocked
+    \item[std::scoped\_lock] RAII to lock and unlock automatically
+    \item[std::unique\_lock] same, but can be released/relocked explicitly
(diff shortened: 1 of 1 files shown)


talk/concurrency.tex
```diff
@@ -233,13 +233,43 @@
...
   \end{exampleblock}
 \end{frame}
 
+\begin{frame}[fragile]
+  \frametitleii{Mutexes and Locks}
+  \begin{block}{Good practice}
+    \begin{itemize}
+      \item Always use \texttt{scoped\_lock} (\cpp11: \texttt{lock\_guard})
+      \item Hold as short as possible, wrap critical section in "\texttt{\string{  \string}}"
+      \item Only if manual control needed, use \texttt{unique\_lock}
```
bernhardmgruber, line 249: Another use case for that is together with `std::condition_variable`.

talk/concurrency.tex
```diff
@@ -233,13 +233,43 @@
...
+  \begin{exampleblock}{}
+    \begin{cppcode*}{gobble=2}
+      std::unique_lock lock1{mutex1, std::try_to_lock};
+      std::unique_lock lock2{mutex2, std::defer_lock};
+      if (lock1) {
+        std::lock(lock2);
+        // ...
+      }
+    \end{cppcode*}
+  \end{exampleblock}
```
bernhardmgruber, line 270: Does this example really help? I know `try_to_lock` and `defer_lock` exist, but how often do you really need them?

Maybe use this additional space to mention `shared_lock`.

talk/concurrency.tex
hageboeck: Good point. We discussed that I'll merge this as the simple version, and there's #62 for the next iteration when we have an advanced course.

talk/concurrency.tex
hageboeck: Also pushed to #62. I'll leave this space blank in the simple version.

PR #61, opened 2021-11-16
Simplify `\frametitle` commands

hageboeck:
On first sight, it's easy to miss that `\frametitleii{}` expands to C++98.
Here, I define
```tex
\frametitlecpp[98, 11, 14, ...]{Title}
```
which should be easier to read.

I can offer to run text replacements on all `frametitlexx`.

This could make things such as [#55](https://github.com/hsf-training/cpluspluscourse/pull/55#issuecomment-948361548) a bit easier.

### DIFF
--- a/talk/C++Course.tex
+++ b/talk/C++Course.tex
@@ -90,14 +90,6 @@
 \newcommand\frametitlecpp[2][98]{
   \frametitle{#2 \hfill \cpp#1}
 }
-% Deprecated:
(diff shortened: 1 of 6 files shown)


hageboeck:
Sorry @sponce that was even faster than I thought. I think it's ready to merge. Without further comments, I'll merge tomorrow.

sponce:
All fine for me. I've almost clicked merge directly, but let's give the evening for others to comment in case and feel free to merge tomorrow indeed.

PR #75, opened 2021-11-26
if constexpr first try

sponce:
Pure rebase of #55 on top of latest master (modulo the fact that small changes were handled in #74)


sponce:
@chavid for info and double check

Issue #69, opened 2021-11-26
Extend slide on optional/any/variant


sponce:
optional/any/variant would really benefit from a short example. the wording on the slides is correct but the students likely won't understand what it's useful for.

theanalyst:
I can take this up (can't assign myself) so commenting here

PR #79, opened 2021-12-16
morestl: add examples for std::optional, any, variant

theanalyst:
Draft WIP #69 PR ...don't have a full tex system currently so will be (ab)using the CI for a bit while I still make up other examples...

### DIFF
--- a/talk/morelanguage/morestl.tex
+++ b/talk/morelanguage/morestl.tex
@@ -76,28 +76,46 @@
 \end{frame}
 
 \begin{frame}[fragile]
-  \frametitlecpp[17]{std::any}
+  \frametitlecpp[17]{std::variant}
   \begin{itemize}
-  \item Indicates the absence of a value with \texttt{std::nullopt}
-  \item Has value semantics (Copy, Move, Compare, stack alloc.)
-  \item Useful in place of pointers where value semantics are intuitive
+  \item Expresses intent that variable holds one of the types
+  \item Makes it easy to implement visitor pattern
   \end{itemize}
+
   \begin{exampleblock}{Code example}
     \begin{cppcode*}{}
-      std::optional<Phone> parse_phone(std::string in)
-      {
-        if (is_valid_phone(in)) {
-          return std::optional<Phone>(in);
-        }
-        return {}; // default constructs std::nullopt
-      }
-      ...
-      auto v = parse_phone(get_input());
-      if (v) {  // alternatively v.is_valid()
-        process_phone(v.value()); // *v is equivalent
-      }
+      using option_t = std::variant<int,float,string>;
+      option_t opt; // default constructs first type
+      std::variant<std::monostate, NonDefault, int, float> myvariant;
+      opt = 100;
+      int ival = std::get<int>(opt); // also std::get<0>(opt)
+      try {
+        float val = std::get<float>(opt) // will throw as opt is int
+      } catch (const std::bad_variant_access& ex) {...}
+
+      // Much easier with get_if returns a ptr or null if not conv.
+      if (const float* val = std::get_if<float>(&opt))
+          std::cout << "option type is f64, value=" << val;
     \end{cppcode*}
   \end{exampleblock}
+
+\end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[17]{Visiting with std::variant}
+    \begin{cppcode*}{A simple visitor}
+      using option_t = std::variant<int,float,string>;
+      struct Visitor {
+        void operator() (int i) { std::cout<< "i32:"<< i;}
+        void operator() (float f) { std::cout<< "f32:"<< f;}
+        void operator() (std::string s) { std::cout<< "str:"<< s;}
+      };
+      void print_opt(option_t opt) {
+        std::visit(Visitor{}, opt);
+      }
+      ...
+      print_opt(100); print_opt(3.14f); print_opt("example");
+    \end{cppcode*}
 \end{frame}
 
 \begin{frame}[fragile]
--- a/talk/morelanguage/morestl.tex
+++ b/talk/morelanguage/morestl.tex
@@ -34,10 +34,9 @@
   \end{itemize}
   \begin{exampleblock}{Code example}
     \begin{cppcode*}{}
-      std::optional<Phone> parse_phone(std::string in)
-      {
+      std::optional<Phone> parse_phone(std::string in) {
         if (is_valid_phone(in)) {
-          return std::optional<Phone>(in);
+          return in; // automatic conv. to std::optional<Phone>(in);
         }
         return {}; // default constructs std::nullopt
       }
--- a/talk/morelanguage/morestl.tex
+++ b/talk/morelanguage/morestl.tex
@@ -119,6 +119,34 @@
 \end{frame}
 
 \begin{frame}[fragile]
+  \frametitlecpp[17]{std::any}
+  \begin{itemize}
+  \item Type safe alternative to void*
+  \item Useful for passing arbitrary values across interfaces that may not know
+    all the types
+  \item \texttt{any\_cast} will only match concrete types, inheritance hierarchy
+    is irrelevant
+  \end{itemize}
+
+  \begin{exampleblock}{Code example}
+    \begin{cppcode*}{}
+      class Orc: public Monster {...};
+
+      void Player::attack(const std::any& e) {
+        if (Monster *m = std::any_cast<Monster>(&e)) {
+          m->take_hit(this->damage);
+        } else if (Orc *o = std::any_cast<Orc>(&e)) {
+          o->take_hit(this->damage*0.5); // Orc will only match this, not its parent
+        } else if (Cat* c = std::any_cast<Cat>(&e)) { // No!
+          c->pet();
+        } ....
+      }
+    \end{cppcode*}
+  \end{exampleblock}
+
+\end{frame}
+
+\begin{frame}[fragile]
   \frametitlecpp[11]{non-member begin and end}
   \begin{alertblock}{The problem in \cpp98}
     STL containers and arrays have different syntax for loop
--- a/talk/morelanguage/morestl.tex
+++ b/talk/morelanguage/morestl.tex
@@ -26,6 +26,56 @@
 \end{frame}
 
 \begin{frame}[fragile]
+  \frametitlecpp[17]{std::optional}
+  \begin{itemize}
+  \item Indicates the absence of a value with \texttt{std::nullopt}
+  \item Has value semantics (Copy, Move, Compare, stack alloc.)
+  \item Useful in place of pointers where value semantics are intuitive
+  \end{itemize}
+  \begin{exampleblock}{Code example}
+    \begin{cppcode*}{}
+      std::optional<Phone> parse_phone(std::string in)
+      {
+        if (is_valid_phone(in)) {
+          return std::optional<Phone>(in);
+        }
+        return {}; // default constructs std::nullopt
+      }
+      ...
+      auto v = parse_phone(get_input());
+      if (v) {  // alternatively v.is_valid()
+        process_phone(v.value()); // *v is equivalent
+      }
+    \end{cppcode*}
+  \end{exampleblock}
+\end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[17]{std::any}
+  \begin{itemize}
+  \item Indicates the absence of a value with \texttt{std::nullopt}
+  \item Has value semantics (Copy, Move, Compare, stack alloc.)
+  \item Useful in place of pointers where value semantics are intuitive
+  \end{itemize}
+  \begin{exampleblock}{Code example}
+    \begin{cppcode*}{}
+      std::optional<Phone> parse_phone(std::string in)
+      {
+        if (is_valid_phone(in)) {
+          return std::optional<Phone>(in);
+        }
+        return {}; // default constructs std::nullopt
+      }
+      ...
+      auto v = parse_phone(get_input());
+      if (v) {  // alternatively v.is_valid()
+        process_phone(v.value()); // *v is equivalent
+      }
+    \end{cppcode*}
+  \end{exampleblock}
+\end{frame}
+
+\begin{frame}[fragile]
   \frametitlecpp[11]{non-member begin and end}
   \begin{alertblock}{The problem in \cpp98}
     STL containers and arrays have different syntax for loop
--- a/talk/morelanguage/morestl.tex
+++ b/talk/morelanguage/morestl.tex
@@ -129,7 +129,7 @@
           m->take_hit(this->damage);
         } else if (Orc *o = std::any_cast<Orc>(&e)) {
           o->take_hit(this->damage*0.5);
-        } else if (Cat* c = std::any_cast<Cat>(&e)) {//No!
+        } else if (Cat *c = std::any_cast<Cat>(&e)) {//No!
           c->pet();
         } ....}
     \end{cppcode*}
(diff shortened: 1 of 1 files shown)


talk/morelanguage/morestl.tex
```diff
@@ -25,6 +25,26 @@ \subsection{More STL}
...
+        return std::optional<Phone>(in);
+      }
+      return std::nullopt; // {} also default constructs to nullopt
+    }
+
+    ...
+    if (auto v = parse\_phone(get\_input());
+        v.is\_valid()) {
+        process\_phone(v.value());
+    }
```
bernhardmgruber, line 44: `std::optional` is contextually convertible to bool, so this should suffice:
```suggestion
    if (auto v = parse\_phone(get\_input())) {
        process\_phone(v.value());
    }
```
Nevertheless, it's good to point out that `has_value() -> bool` also exists.

talk/morelanguage/morestl.tex
```diff
@@ -25,6 +25,26 @@ \subsection{More STL}
...
+\begin{frame}[fragile]
+  \frametitlecpp[17]{std::optional}
+  \scriptsize
+  \begin{cppcode}
+    std::optional<Phone> parse\_phone(std::string\_view in)
+    {
+      if (is\_valid\_phone(in)) {
+        return std::optional<Phone>(in);
+      }
+      return std::nullopt; // {} also default constructs to nullopt
```
sponce, line 42: I would invert code and comment
```suggestion
      return {}; // default constructs to std::nullopt
```

talk/morelanguage/morestl.tex
sponce: Agreed that the contextual conversion to `bool` should be demonstrated. But again I would avoid mixing it with other features that may not yet been assimilated, here the usage of init-statement in particular and the fact that `operator=` returns its result. So let's maybe simplify to 
```cpp
  auto v = parse_phone(get_input());
  if (v) {
     process_phone(v.value());  // or *v, equivalent to v.value()
  }
```

talk/morelanguage/morestl.tex
```diff
@@ -25,6 +25,26 @@ \subsection{More STL}
   \end{block}
 \end{frame}
 
+\begin{frame}[fragile]
+  \frametitlecpp[17]{std::optional}
+  \scriptsize
+  \begin{cppcode}
+    std::optional<Phone> parse\_phone(std::string\_view in)
```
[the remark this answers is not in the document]
theanalyst: Makes sense, will use `std::string` for now

talk/morelanguage/morestl.tex
```diff
@@ -25,6 +25,26 @@ \subsection{More STL}
...
+    {
+      if (is\_valid\_phone(in)) {
```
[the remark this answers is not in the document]
theanalyst: yes, just checked the output to realize... fixing

talk/morelanguage/morestl.tex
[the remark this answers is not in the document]
theanalyst: Might as well introduce `std::string_view` in the More STL subsection if everyone agrees :)

talk/morelanguage/morestl.tex
```diff
@@ -25,6 +25,56 @@ \subsection{More STL}
...
+  \item Indicates the absence of a value with \texttt{std::nullopt}
+  \item Has value semantics (Copy, Move, Compare, stack alloc.)
+  \item Useful in place of pointers where value semantics are intuitive
+  \end{itemize}
+  \begin{exampleblock}{Code example}
+    \begin{cppcode*}{}
+      std::optional<Phone> parse_phone(std::string in)
+      {
+        if (is_valid_phone(in)) {
+          return std::optional<Phone>(in);
```
sponce, line 40:
```suggestion
          return in ; // automatic conversion to std::optional<Phone>
```

talk/morelanguage/morestl.tex
```diff
@@ -25,6 +25,56 @@ \subsection{More STL}
...
+  \frametitlecpp[17]{std::optional}
+  \begin{itemize}
+  \item Indicates the absence of a value with \texttt{std::nullopt}
+  \item Has value semantics (Copy, Move, Compare, stack alloc.)
+  \item Useful in place of pointers where value semantics are intuitive
+  \end{itemize}
+  \begin{exampleblock}{Code example}
+    \begin{cppcode*}{}
+      std::optional<Phone> parse_phone(std::string in)
+      {
```
sponce, line 38: Let's gain a line but putting the brace at the end of previous one

talk/morelanguage/morestl.tex
```diff
@@ -25,6 +25,56 @@ \subsection{More STL}
...
+      auto v = parse_phone(get_input());
+      if (v) {  // alternatively v.is_valid()
+        process_phone(v.value()); // *v is equivalent
+      }
+    \end{cppcode*}
+  \end{exampleblock}
+\end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[17]{std::any}
```
sponce, line 54: I suppose you will fix this slide, which is for the moment a copy of the previous one ?

theanalyst: Yes indeed sorry didn't mean to commit that... was planning on adding similar content for std::any and variant

talk/morelanguage/morestl.tex
bernhardmgruber: You can do as you like. I still prefer to have
```c++
  if (auto v = parse_phone(get_input())) {
     process_phone(v.value());  // or *v, equivalent to v.value()
  }
```
on the slide somewhere, because this is what I would like the students to remember and write in their code bases.

theanalyst:
Added slides on `std::variant` and `std::any` now; also a brief one on `std::string_view`. I think this is ready for review now, happy to change anything.

sponce:
Now that the different slides are written, I feel like the original one should be dropped and the content distributed into the different sections. It feels odd to start with it, then switch to string_view and comes back to the 3 items, one by one.

talk/morelanguage/morestl.tex
```diff
@@ -25,6 +25,121 @@ \subsection{More STL}
...
+  \begin{itemize}
+  \item Expresses intent that variable holds one of the types
+  \item Makes it easy to implement visitor pattern
+  \end{itemize}
+
+  \begin{exampleblock}{Code example}
+    \begin{cppcode*}{}
+      using option_t = std::variant<int,float,string>;
+      option_t opt; // default constructs first type
+      std::variant<std::monostate, NonDefault, int> myvar;
```
sponce, line 84: Is this line a remnant of a previous piece of code ?
```suggestion
```

talk/morelanguage/morestl.tex
```diff
@@ -25,6 +25,121 @@ \subsection{More STL}
...
+      opt = 100;
```
sponce, line 85: why not directly
```cpp
option_t opt{100};
```

talk/morelanguage/morestl.tex
```diff
@@ -25,6 +25,121 @@ \subsection{More STL}
...
+      int ival = std::get<int>(opt);//also std::get<0>(opt)
+      try {
+        float val = std::get<float>(opt) // will throw...
+      } catch (const std::bad_variant_access& ex) {...}
+
+      // Much easier with get_if; returns a ptr or null
+      if (const float* val = std::get_if<float>(&opt))
+          std::cout<<"option type is f64, value="<< val;
```
sponce, line 93: Not sure I like to reintroduce pointers here. On top you're printing the pointer not the value.

talk/morelanguage/morestl.tex
```diff
@@ -25,6 +25,121 @@ \subsection{More STL}
...
+    \end{cppcode*}
+  \end{exampleblock}
+
+\end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[17]{Visiting with std::variant}
```
sponce, line 100: Nice slide, would maybe benefit also of lambdas syntax and lambda inheritance. We would then have to switch the lambda part and the MoreSTL one. What do people think ?

talk/morelanguage/morestl.tex
theanalyst: I kind of wanted to say that variant requires the first type to be default constructable, (hence the monostate line) but probably not coherent/making sense here..

talk/morelanguage/morestl.tex
theanalyst: sure makes sense

theanalyst:
> Now that the different slides are written, I feel like the original one should be dropped and the content distributed into the different sections. It feels odd to start with it, then switch to string_view and comes back to the 3 items, one by one.

This makes sense, also makes it easy slide estate planning wise so I could mostly go for one slide explaining the item and the next one with the program example

talk/morelanguage/morestl.tex
theanalyst: I kind of wanted to mention alternatives to simply try catch using the get api... but we anyway introduce the visitor pattern in the next slide, so this can be skipped

talk/morelanguage/morestl.tex
theanalyst: So my only comment on this is since at first glance this is an assignment instead of an expression, (but if RHS is null the if condition would fail), would this be intuitive to students?

talk/morelanguage/morestl.tex
theanalyst: Sure, so we can add the lambda visitor after the lambda examples, saying the earlier written visitor can also be written this way. What do you think?

sponce: Yes, that's even better indeed !

talk/morelanguage/morestl.tex
hageboeck, line 93: Here is an attempt to shorten code. Hopefully, this is easier to understand for novices.
Feel free to use only in part or not at all.
```suggestion
      std::variant<int,float,string> opt{100}; // Now holding int
      int ival = std::get<int>(opt); //also std::get<0>(opt)
      try {
        float val = std::get<float>(opt) // will throw...
      } catch (const std::bad_variant_access& ex) {...}

      if (std::holds_alternative<float>(opt))
          std::cout<<"option type is f32, value="<< std::get<float>(opt);
```

talk/morelanguage/morestl.tex
```diff
@@ -25,6 +25,121 @@ \subsection{More STL}
...
+  \begin{exampleblock}{Code example}
+    \begin{cppcode*}{}
+      class Orc: public Monster {...};
+
+      void Player::attack(const std::any& e) {
+        if (Monster *m = std::any_cast<Monster>(&e)) {
+          m->take_hit(this->damage);
+        } else if (Orc *o = std::any_cast<Orc>(&e)) {
+          o->take_hit(this->damage*0.5);
+        } else if (Cat* c = std::any_cast<Cat>(&e)) {//No!
```
hageboeck, line 135:
```suggestion
        } else if (Cat *c = std::any_cast<Cat>(&e)) {//No!
```

talk/morelanguage/morestl.tex
```diff
@@ -25,6 +25,121 @@ \subsection{More STL}
...
+\end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[17]{std::any}
+  \begin{itemize}
+  \item Type safe alternative to void*
+  \item Useful for passing arbitrary values across interfaces that may not know
+    all the types
+  \item \texttt{any\_cast} will only match concrete types, inheritance hierarchy
+    is irrelevant
```
hageboeck, line 123: Should we mention (or use colours) the fact (surprising for some people) that you cast like:
```c++
int *i = std::any_cast<int>(&any);
// and NOT
int *i = std::any_cast<int*>(&any);
```

sponce: let's not mention it for now and see whether it's perturbing students or not

sponce:
@theanalyst  This is almost ready to be merged, but if you do not mind, I would go for dropping the introduction slide and merge it in the others. I can do it myself if you agree

talk/morelanguage/morestl.tex
```diff
@@ -25,6 +25,118 @@ \subsection{More STL}
...
+    \begin{cppcode*}{}
+      std::variant<int,float,string> opt{100}; // Now holding int
+      int ival = std::get<int>(opt); //also std::get<0>(opt)
+      try {
+        float val = std::get<float>(opt) // will throw...
+      } catch (const std::bad_variant_access& ex) {...}
+
+      // Or check the type before accessing it
+      if (std::holds_alternative<float>(opt))
+          std::cout<<"option type is fp32, value="<< std::get<float>(opt);
```
hageboeck, line 90: The comments / code leak out of the `\exampleblock`:
```suggestion
      std::variant<int,float,string> opt{100}; // Now int
      int ival = std::get<int>(opt); //Or std::get<0>(opt)
      try {
        float val = std::get<float>(opt) // will throw...
      } catch (const std::bad_variant_access& ex) {...}

      // Or check the type before accessing it
      if (std::holds_alternative<float>(opt))
          std::cout << "option type is fp32, value="
              << std::get<float>(opt);
```

talk/morelanguage/morestl.tex
```diff
@@ -25,6 +25,118 @@ \subsection{More STL}
...
+  \frametitlecpp[17]{Visiting with std::variant}
+    \begin{cppcode*}{}
+      using option_t = std::variant<int,float,string>;
+      struct Visitor {
+        void operator() (int i) { std::cout<< "i32:"<< i;}
+        void operator() (float f) { std::cout<< "f32:"<< f;}
+        void operator() (string s) { std::cout<< "s:"<< s;}
+      };
+      void print_opt(option_t opt) {
+        std::visit(Visitor{}, opt);
```
hageboeck, line 106:
```suggestion
        std::visit(Visitor{}, opt);
        std::cout << '\n';
```

hageboeck (commented):
I have one formatting thing and one optional line break. Otherwise, the output of the program would look a bit complicated.

theanalyst:
> @theanalyst This is almost ready to be merged, but if you do not mind, I would go for dropping the introduction slide and merge it in the others. I can do it myself if you agree

Yes sure go ahead, sorry things got a bit too busy after the vacation

sponce:
Ok, will do. I was also busy until now, so starting only today. One question I have though. The code presented for std::any looks wrong to me :
```cpp
 if (Monster *m = std::any_cast<Monster>(&e)) {
          m->take_hit(this->damage);
        } else if (Orc *o = std::any_cast<Orc>(&e)) {
          o->take_hit(this->damage*0.5);
        } else if (Cat *c = std::any_cast<Cat>(&e)) {//No!
```
it assumes `std::any_cast` will return `nullptr`, but it will throw `std::bad_any_cast`, not return

hageboeck:
> it assumes `std::any_cast` will return `nullptr`, but it will throw `std::bad_any_cast`, not return

Actually, only the reference and universal reference versions will throw @sponce. The pointer versions do return `nullptr`. [cppreference](https://en.cppreference.com/w/cpp/utility/any/any_cast).
But what I find very confusing on first look is that the template argument is the type `<Orc>` whereas the argument of the cast is a pointer `(&e)`.
This is different from what you normally have, e.g. `static_cast<Type*>(&e)`. This is what we should discuss in detail. If we both get confused on first look (in slightly different directions, admittedly), learners might be confused as well.

sponce:
On now I've seen the little "(1-3)" in the cpp reference ! Thanks @hageboeck for being more awake than I was.
I find this extremely confusing...

I anyway would not take an example using pointers. After all we want to teach people to drop them everywhere.

sponce:
So I've created a new version of this one in #89 (could not push to @theanalyst' s branch). I've changed the examples for any and if constexpr. @hageboeck if you could review, would be appreciated

PR #81, opened 2021-12-17
New exceptions section, recommended modern use

graeme-a-stewart:
New exceptions section with more illustrative examples
of recommended use. This includes some discussion of
what one should use exceptions for, what one should not
use them for. This includes some performance considerations.

All of the deprecated `throw()` syntax is removed, and
the slide on the more current `nothrow` is expanded.

A few additional LaTeX ephemeral files are added to
.gitignore (arise when processing with MiKTeX).

### DIFF
--- a/talk/morelanguage/exceptions.tex
+++ b/talk/morelanguage/exceptions.tex
@@ -4,185 +4,166 @@
   \frametitlecpp[98]{Exceptions}
   \begin{block}{The concept}
     \begin{itemize}
-    \item exceptional event breaking linearity of the code
-    \item will be handled in dedicated place
+      \item to handle \textit{exceptional} events that cannot be foreseen
+      \item to cleanly jump to the place where the error can be handled
     \end{itemize}
   \end{block}
-  \begin{block}{Practically}
+  \begin{block}{In practice}
     \begin{itemize}
-    \item you can throw any object with {\it throw}
-    \item you handle them using {\it try ... catch} blocks
+      \item add an exception handling block with {\it try ... catch}
+      \item throw an exception within it using {\it throw}
     \end{itemize}
   \end{block}
-  \begin{cppcode}
+  \begin{cppcode*}{fontsize=\scriptsize}
+    #include <stdexcept>
+    ...
     try {
-      if (0 == name) {
-        throw std::string("Expected non empty name");
+      if (0 == divisor) {
+        throw illegal_argument("Cannot divide by zero");
       }
-      printf("%s\n", name);
-    } catch (std::string& e) {
-      printf("empty name found\n");
+      ... // Normal processing
+    } catch (const invalid_argument& e) {
+      cerr << e.what() << endl;
     }
-  \end{cppcode}
+  \end{cppcode*}
 \end{frame}
 
 \begin{frame}[fragile]
   \frametitlecpp[98]{Exceptions}
-  \begin{block}{Rules}
+  \begin{block}{Rules and Advice}
     \begin{itemize}
-    \item exception will skip all code until next {\it catch}
-    \begin{itemize}
-      \item still destructors are called when exiting scopes
-      \item but your own cleanup may not be
+      \item any object can be thrown, but best to use those in \texttt{stdexcept}
+      \begin{itemize}
+        \item define your own subclass of \texttt{std::exception} if needed
+      \end{itemize}
+      \item exceptions will be caught if the object type of \textit{catch} matches
+      \begin{itemize}
+          \item if nothing catches an exception then \texttt{std::terminate} is called
+          \item throw exceptions by value, catch them by (const) reference
+      \end{itemize}
+      \item all objects on the stack between the \textit{throw} and the \textit{catch} have their destructors called
+      \begin{itemize}
+        \item this should give you a clean release of intermediate resources
+        \item make sure you are using the RAII idiom in your own classes
+      \end{itemize}
+      \item use exceptions for runtime errors like bad inputs, file not found, DB connection, \ldots
+      \item \textit{don't} use exceptions for logic errors in your code (use \texttt{assert})
     \end{itemize}
-    \item {\it catch} is selective on the exception type
+  \end{block}
+\end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[98]{Exceptions}
+  \begin{block}{An more illustrative example}
+    \begin{itemize}
+      \item exceptions are very powerful when there are many calls / lines of code between the error and where the error is handled
+      \item they can also rather cleanly handle different types of errors
+      \item \textit{try ... catch} statements can also be nested
     \end{itemize}
   \end{block}
   \begin{multicols}{2}
     \begin{cppcode*}{fontsize=\scriptsize,gobble=2}
-      struct ZeroDivide {};
-
-      int divide(int a, int b) {
-        if (0 == b) {
-          throw ZeroDivide();
+      try {
+        for (const &file : files) {
+          try {
+            process_file(file);
+          }
+          catch (const file_bad &e) {
+            ... // Skips one file
+          }
         }
-        return a/b;
+      } catch (const db_bad &e) {
+        ... // This exception
+            // aborts everything
       }
     \end{cppcode*}
     \columnbreak
-    \begin{cppcode*}{fontsize=\scriptsize,gobble=2,firstnumber=9}
-      int func(char* value) {
+    \begin{cppcode*}{fontsize=\scriptsize,gobble=2}
+      int process_file(const fileobj file) {
+        if (fh = open_file(file)) {
+          throw
+            file_bad(file.error.reason());
+        }
+        while (!fh) {
+          line = read_line(fh);
+          add_line_to_db(line); // Can throw
+                                // db_bad
+        }
+        return 0;
+      }
+    \end{cppcode*}
+  \end{multicols}
+\end{frame}
+
+
+\begin{frame}[fragile]
+  \frametitlecpp[98]{Error Handling and Exceptions}
+  \begin{block}{}
+    \begin{itemize}
+      \item exceptions have little cost if no exception is thrown
+      \begin{itemize}
+        \item a throw usually costs about as much as a function call
+        \item so minimal if you are using exceptions for \textit{exceptional} events
+        \item thus their use is generally recommended
+      \end{itemize}
+      \item however, in a tight, performance critical loop, where the error is handled close to the place it happens a C-style error code might be better
+      \begin{itemize}
+        \item But profile if you really want to find out!
+      \end{itemize}
+   \end{itemize}
+  \end{block}
+  \begin{multicols}{2}
+    \begin{cppcode*}{fontsize=\tiny,gobble=2}
+      // try ... catch in hot loop
+      //   might impact performance
+      for (const &item: list) {
         try {
-          errno = 0;
-          long l = strtol(value,0,10);
-          if (errno) {
-            throw string("Bad Value");
-          }
-          divide(100, l);
-        } catch (string& e) {
-          printf("%s\n", e.c_str());
-        } catch (ZeroDivide e2) {
-          printf("Division error\n");
+          check_item(item); // Can throw bad_item
+          ... // process item
+        }
+        catch (bad_item &e) {
+          ... // ignore this item, but keep going
+        }
+      }
+    \end{cppcode*}
+    \columnbreak
+    \begin{cppcode*}{fontsize=\tiny,gobble=2}
+      // For this code logic C-style error codes
+      //   may be better
+      for (const &item: list) {
+        auto err_code = check_item(item);
+        if (!err_code) {
+          ... // process item
+        } else {
+          ... // warn and carry on
         }
       }
     \end{cppcode*}
   \end{multicols}
 \end{frame}
 
-\begin{frame}[fragile]
-  \frametitlecpp[98]{Controlling exceptions \hfill \deprecated/\removed}
-  \begin{block}{Declaring expected exceptions}
-    \begin{itemize}
-    \item each function can declare a set of expected exceptions
-    \item using the {\it throw} statement in its declaration
-    \item other exceptions won't exit the scope of the function
-      \begin{itemize}
-      \item instead, the {\it unexpected} handler is called
-      \item by default, it terminates the program
-      \end{itemize}
-    \end{itemize}
-  \end{block}
-  \pause
-  \begin{cppcode}
-    int func(int a) throw(int) {
-      if (0 == a) {
-        throw 2;  // ok, goes out
-      } else {
-        throw "hello"; // std::unexpected called
-      }
-    }
-  \end{cppcode}
-\end{frame}
 
 \begin{frame}[fragile]
-  \frametitlecpp[98]{Controlling exceptions \hfill \deprecated/\removed}
-  \begin{block}{Good to know}
+  \frametitlecpp[11]{noexcept}
+  \begin{block}{noexcept}
     \begin{itemize}
-    \item The check is done at runtime, not at compile time
-      \begin{itemize}
-      \item unlike Java
-      \end{itemize}
-    \item When the {\it throw} clause is absent, any exception can go out
-    \item To block all exceptions, use {\it throw()}
-    \end{itemize}
-  \end{block}
-  \pause
-  \begin{cppcode}
-    int func(int a) {
-      // any exception can go out
-    }
-    int otherfunc(int a) throw() {
-      // no exception can go out
-    }
-  \end{cppcode}
-\end{frame}
-
-\begin{frame}[fragile]
-  \frametitlecpp[11]{Removal of \cpp98 exceptions}
-  \begin{block}{}
-    After a lot of thinking and experiencing, the conclusions of the community on exception handling are :
-    \begin{itemize}
-    \item Never write an exception specification
-    \item Except possibly an empty one
-    \end{itemize}
-  \end{block}
-  \pause
-  \begin{alertblock}{Some of the reasons}
-    \begin{itemize}
-    \item throw specification is runtime only
-      \begin{itemize}
-      \item does not allow compiler optimizations
-      \item on the contrary forces extra checks
-      \item generally terminates your program if violated
-      \end{itemize}
-    \item throw specification clashes with templates
-      \begin{itemize}
-      \item one cannot ``template'' the throw clause
-      \end{itemize}
-    \end{itemize}
-  \end{alertblock}
-\end{frame}
-
-\begin{frame}[fragile]
-  \frametitlecpp[11]{What remains}
-  \begin{block}{throw is dead}
-    \begin{itemize}
-    \item dynamic exception specification are deprecated
-    \item even throw() (no exceptions)
-    \end{itemize}
-  \end{block}
-  \pause
-  \begin{exampleblock}{long live noexcept}
-    \begin{itemize}
-    \item noexcept a somehow equivalent to throw()
-    \item but is checked at compile time
-    \item so allows compiler optimizations
-    \end{itemize}
-  \end{exampleblock}
-\end{frame}
-
-\begin{frame}[fragile]
-  \frametitlecpp[11]{Full power of noexcept}
-  \begin{block}{3 uses of noexcept}
-    \begin{itemize}
-    \item standalone
-      \begin{cppcode*}{gobble=2, linenos=false}
+      \item specifying the \textit{noexcept} operator states that a function is guaranteed not to throw an exception
+      \begin{cppcode*}{fontsize=\scriptsize,gobble=2,linenos=false}
         int f() noexcept;
       \end{cppcode*}
-    \item as an expression saying whether exceptions can be sent
-      \begin{cppcode*}{gobble=2, linenos=false}
-        int f() noexcept(sizeof(long) == 8);
-      \end{cppcode*}
-    \item as an operator to know whether a function launches exceptions
-      \begin{cppcode*}{gobble=2, linenos=false}
-        constexpr bool callCannotThrow = noexcept(f());
-        if constexpr (callCannotThrow) { ... }
-      \end{cppcode*}
-    \item and you can combine them
-      \begin{cppcode*}{gobble=2, linenos=false}
-        template <typename T> void foo()
-             noexcept(noexcept(T())) {}
-      \end{cppcode*}
-   \end{itemize}
+      \begin{itemize}
+        \item either no exceptions will ever be thrown or any they are handled internally
+        \item checked at compile time, so it allows the compiler to optimise around that knowledge
+      \end{itemize}
+      \item \textit{noexcept(expression)} evaluates the expression and, if \textit{true} the function won't throw
+        \begin{cppcode*}{fontsize=\scriptsize,gobble=2,linenos=false}
+          int safe_if_long_is_8_bytes() noexcept(sizeof(long)==8);
+        \end{cppcode*}
+      \item These uses can be nested, e.g.,
+        \begin{cppcode*}{fontsize=\scriptsize,gobble=2,linenos=false}
+          int ok_if_f_and_g_ok() noexcept(noexcept(f()) && noexcept(g()));
+        \end{cppcode*}
+      \item If you do use \textit{noexcept}, pay attention to destructors and don't allow exceptions to escape (or ever let resources leak)
+    \end{itemize}
   \end{block}
 \end{frame}
--- a/talk/morelanguage/exceptions.tex
+++ b/talk/morelanguage/exceptions.tex
@@ -10,7 +10,7 @@
   \end{block}
   \begin{block}{In practice}
     \begin{itemize}
-      \item add an exception handling block with {\it try ... catch} when exceptions are possible
+      \item add an exception handling block with {\it try ... catch} when exceptions are possible \textit{and can be handled}
       \item throw an exception using {\it throw} when a function cannot proceed (or recover internally)
     \end{itemize}
   \end{block}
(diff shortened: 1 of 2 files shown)


graeme-a-stewart:
Just a few additional comments from my side, that might be worth discussion:

- I based most of this on the nice guide here <https://docs.microsoft.com/en-us/cpp/cpp/errors-and-exception-handling-modern-cpp?view=msvc-170>
- I dropped all of the old discussion on `throw()`, that I think should not be mentioned at all (I don't even think it was widely used, at least not in code I ever worked with)
- The *more illustrative example* (section slide 3) I worry is a bit too complex, but I was trying to get something that illustrated some of the more realistic uses and important points (I think "look with the eyes of a beginner", is it too much of a leap from the trivial example on slide 1?)

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,166 @@
   \frametitlecpp[98]{Exceptions}
   \begin{block}{The concept}
     \begin{itemize}
-    \item exceptional event breaking linearity of the code
-    \item will be handled in dedicated place
+      \item to handle \textit{exceptional} events that cannot be foreseen
+      \item to cleanly jump to the place where the error can be handled
```
bernhardmgruber, line 8:
```suggestion
      \item to cleanly jump to a place where the error can be handled
```
I like to propose this small distinction, because in many cases, we don't know where an error will be handled.

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,166 @@
...
+      if (0 == divisor) {
+        throw illegal_argument("Cannot divide by zero");
       }
-      printf("%s\n", name);
-    } catch (std::string& e) {
-      printf("empty name found\n");
+      ... // Normal processing
+    } catch (const invalid_argument& e) {
+      cerr << e.what() << endl;
     }
```
bernhardmgruber, line 27: I would be alarmed if a try/catch contained its own throw statement in the same function. This should be handled with normal control flow constructs instead. Could you try to phrase the example in a way that the try/catch and throw are in different functions?

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,166 @@
...
     \begin{itemize}
-    \item exception will skip all code until next {\it catch}
-    \begin{itemize}
-      \item still destructors are called when exiting scopes
-      \item but your own cleanup may not be
+      \item any object can be thrown, but best to use those in \texttt{stdexcept}
+      \begin{itemize}
+        \item define your own subclass of \texttt{std::exception} if needed
+      \end{itemize}
+      \item exceptions will be caught if the object type of \textit{catch} matches
```
bernhardmgruber, line 39:
```suggestion
      \item exceptions will be caught if the object type in the \textit{catch clause} matches
```
Otherwise I wonder what `catch` refers to?

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,166 @@
...
+      \begin{itemize}
+          \item if nothing catches an exception then \texttt{std::terminate} is called
+          \item throw exceptions by value, catch them by (const) reference
+      \end{itemize}
+      \item all objects on the stack between the \textit{throw} and the \textit{catch} have their destructors called
+      \begin{itemize}
+        \item this should give you a clean release of intermediate resources
+        \item make sure you are using the RAII idiom in your own classes
+      \end{itemize}
+      \item use exceptions for runtime errors like bad inputs, file not found, DB connection, \ldots
```
bernhardmgruber, line 49:
```suggestion
      \item use exceptions for unlikely runtime errors like bad inputs, file not found, DB connection, \ldots
```
You could still check whether a config file is present or not and a "file not found" could be part of your normal business logic. I would recommend a check for file existence there instead of catching an exception. Exceptions should not be part of any normal business logic.

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,166 @@
...
-  \pause
-  \begin{alertblock}{Some of the reasons}
-    \begin{itemize}
-    \item throw specification is runtime only
+      \item exceptions have little cost if no exception is thrown
       \begin{itemize}
-      \item does not allow compiler optimizations
-      \item on the contrary forces extra checks
-      \item generally terminates your program if violated
+        \item a throw usually costs about as much as a function call
```
bernhardmgruber, line 105: That depends on a lot of factors. I cannot find the paper on performance of various kind of error handling, but I remember exceptions where crazy expensive if you throw from a deeper call hierarchy or if you throw polymorphic values. But I cannot request you to change your wording until I find the numbers :/

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,166 @@
...
+        \item so minimal if you are using exceptions for \textit{exceptional} events
+        \item thus their use is generally recommended
       \end{itemize}
-    \item throw specification clashes with templates
+      \item however, in a tight, performance critical loop, where the error is handled close to the place it happens a C-style error code might be better
```
bernhardmgruber, line 109: I would not recommend solely a C-style error code. We have `boost::outcome`, `std::expected`, `std::system_error` or just a plain sentinal value (e.g. returning -1 if the normal return values are expected to be positive). Also C-style error code could imply `errno` here, which is a `thread_local` variable and that is also expensive.

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,166 @@
...
+
 \begin{frame}[fragile]
-  \frametitlecpp[11]{Full power of noexcept}
-  \begin{block}{3 uses of noexcept}
+  \frametitlecpp[11]{noexcept}
+  \begin{block}{noexcept}
     \begin{itemize}
-    \item standalone
-      \begin{cppcode*}{gobble=2, linenos=false}
+      \item specifying the \textit{noexcept} operator states that a function is guaranteed not to throw an exception
```
bernhardmgruber, line 150:
```suggestion
      \item specifying the \textit{noexcept} specifier states that a function is guaranteed not to throw an exception
```
The `noexcept` operator (written the same way) is used to query that property from a function, like `bool b = noexcept(f(2, 54));`

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,166 @@
...
-        if constexpr (callCannotThrow) { ... }
-      \end{cppcode*}
-    \item and you can combine them
-      \begin{cppcode*}{gobble=2, linenos=false}
-        template <typename T> void foo()
-             noexcept(noexcept(T())) {}
-      \end{cppcode*}
-   \end{itemize}
+      \begin{itemize}
+        \item either no exceptions will ever be thrown or any they are handled internally
```
bernhardmgruber, line 155:
```suggestion
        \item either no exceptions will ever be thrown or they are handled internally
```
or
```suggestion
        \item no exceptions will ever escape from the function
```

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,166 @@
...
+        \item checked at compile time, so it allows the compiler to optimise around that knowledge
```
bernhardmgruber, line 156: Or pessimize. If you call a not noexcept function inside a noexcept function, the compiler needs to generate a try/catch block with a call to terminate. That is slower than if all functions had no noexcept specifier.

Therefore: noexcept should only appear on leaf functions.

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,166 @@
...
+      \end{itemize}
+      \item \textit{noexcept(expression)} evaluates the expression and, if \textit{true} the function won't throw
```
bernhardmgruber, line 158: That is confusing:
`bool b = noexcept(expression)` does NOT evaluate the expression. In fact, it is an evaluated context.
`void f() noexcept(expression)` DOES evaluate the expression.
You can also combine them: `void f() noexcept(noexcept(expression))` which does NOT evaluate expression.
Please make a better distinction whether you are talking about the noexcept operator or specifier.

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,166 @@
...
+        \begin{cppcode*}{fontsize=\scriptsize,gobble=2,linenos=false}
+          int safe_if_long_is_8_bytes() noexcept(sizeof(long)==8);
+        \end{cppcode*}
+      \item These uses can be nested, e.g.,
+        \begin{cppcode*}{fontsize=\scriptsize,gobble=2,linenos=false}
+          int ok_if_f_and_g_ok() noexcept(noexcept(f()) && noexcept(g()));
+        \end{cppcode*}
+      \item If you do use \textit{noexcept}, pay attention to destructors and don't allow exceptions to escape (or ever let resources leak)
```
bernhardmgruber, line 166: Hmm. Destructors are implicitely marked `noexcept`, so why the special attention to them? You cannot even implement a throwing destructor (it will `std::terminate`) unless you explicitly mark it as `noexcept(false)`. Did it once in my life, caused a lot of troubles. Don't do it.

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,166 @@
...
+      \item to cleanly jump to the place where the error can be handled
     \end{itemize}
   \end{block}
-  \begin{block}{Practically}
+  \begin{block}{In practice}
     \begin{itemize}
-    \item you can throw any object with {\it throw}
-    \item you handle them using {\it try ... catch} blocks
+      \item add an exception handling block with {\it try ... catch}
+      \item throw an exception within it using {\it throw}
```
hageboeck, line 14: Actually, don't throw in that block, as Bernhard said. So maybe:
```suggestion
      \item add an exception handling block with {\it try ... catch} when exceptions are expected
      \item throw an exception using {\it throw} when a function cannot proceed with the intended task
```

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,166 @@
...
+      \item any object can be thrown, but best to use those in \texttt{stdexcept}
+      \begin{itemize}
+        \item define your own subclass of \texttt{std::exception} if needed
+      \end{itemize}
+      \item exceptions will be caught if the object type of \textit{catch} matches
+      \begin{itemize}
+          \item if nothing catches an exception then \texttt{std::terminate} is called
+          \item throw exceptions by value, catch them by (const) reference
+      \end{itemize}
+      \item all objects on the stack between the \textit{throw} and the \textit{catch} have their destructors called
```
hageboeck, line 44:
```suggestion
      \item all objects on the stack between the \textit{throw} and the \textit{catch} are destructed automatically
```

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,166 @@
...
+      \begin{itemize}
+        \item this should give you a clean release of intermediate resources
+        \item make sure you are using the RAII idiom in your own classes
+      \end{itemize}
+      \item use exceptions for runtime errors like bad inputs, file not found, DB connection, \ldots
+      \item \textit{don't} use exceptions for logic errors in your code (use \texttt{assert})
```
hageboeck, line 50: `assert` are useless in release builds, so you either get wrong results or a crash.
And I think there's logic errors that would deserve some kind of clean termination. At least an exception has a `what()` message that you could inspect.
You can also print an error message and `abort`, so I think we should find some general advice we can give to users.

talk/morelanguage/exceptions.tex
hageboeck: Throw might require you to unwind of stuff and tear it down properly. The statement about the low cost when they are rare likely holds, though.
Just don't abuse them to pass information around.

talk/morelanguage/exceptions.tex
hageboeck, line 109:
```suggestion
      \item however, in a tight, performance-critical loop, where the error is handled close to the place it happens a C-style error code might be better
```

talk/morelanguage/exceptions.tex
graeme-a-stewart: I'm not convinced - although it's a minor point the programmer should know where the exception is going to be caught as part of the code design (and the compiler definitely knows 😄)

talk/morelanguage/exceptions.tex
graeme-a-stewart: Fair enough - it was a rework of the original example, not enough I accept!

talk/morelanguage/exceptions.tex
graeme-a-stewart: Based on : https://docs.microsoft.com/en-us/cpp/cpp/errors-and-exception-handling-modern-cpp?view=msvc-170#exceptions-and-performance - could be compiler specific of course. I found a few stack overflow articles claiming the same, that the cost is throwing the exception, not the try ... catch. Throwing unwinds the stack, but that would be just as expensive when passed up through error values as done with an exception (and then also far more fragile).

talk/morelanguage/exceptions.tex
bernhardmgruber: Not really. When I throw in my library, I have no clue which other library/application includes me as a third-party dependency and whether they can handle that. I can just document and hope for the best.
Do you know what exceptions or whether `boost::filesystem::exists()` can throw?  Do the authors of Boost.Filesystem now whether your code will catch or not, or what types it tries to catch? Sadly, not everybody inherits from `std::exception` too. We live in a world where software is build from many dependencies that we don't control. They could throw anything (even int or std::string).

And certainly the compiler does not know in many cases, e.g. if you load a shared object that you pulled form the internet or that was part of a commercial/closed-source software.

talk/morelanguage/exceptions.tex
graeme-a-stewart: I am thinking more of cases that crop up in HEP where the files *should* be there, e.g., they are registered in the file catalogue of the experiment. Thus not being able to access them is indeed *exceptional* (like the mass storage system is down). I added "unexpectedly" at least.

If you can think of better examples please suggest them!

talk/morelanguage/exceptions.tex
graeme-a-stewart: Rephrased this to be a bit clearer that the throw is ~like an *extra* call and that the stack always needs to be unwound

talk/morelanguage/exceptions.tex
graeme-a-stewart: Yes, good suggestion - I have changed the advice generally to suggest this for leaf functions

talk/morelanguage/exceptions.tex
graeme-a-stewart: Yep, honestly I'd say this is rather an advanced topic (I was indeed confused). I have tried to simplify this to essentials that are useful now.

talk/morelanguage/exceptions.tex
graeme-a-stewart: Ditto for comment above.

talk/morelanguage/exceptions.tex
graeme-a-stewart: So a logic error means that you made a programming mistake so this is, by definiton, unrecovereable - the program needs fixed. All an exception does here is beautify things. It's also doubtful here that `e.what()` can tell you anything useful and you now lost the stack that could have been useful (I think even if you re-throw you lost some context?).

I still think `assert()` is the useful way to test all of the possible outcomes, although it's sometimes hard to anticipate all cases, that's true...

talk/morelanguage/exceptions.tex
graeme-a-stewart: I tried to be a bit more precise re. the cost and that stack unwind happens always - I hope we got the message across that you don't abuse exceptions in C++ as normal controlflow (don't use EAFP, unlike Python).

talk/morelanguage/exceptions.tex
graeme-a-stewart: There is documentation, but anyway, this is a minor issue so I changed it.

graeme-a-stewart:
Thanks a lot for the reviews @bernhardmgruber and @hageboeck. I think I addressed most things, but please cast your eye over again.

Should we add something on `catch(...)`, thinking about @bernhardmgruber's comment that throwing `std::string` (and similar) still happens?

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,172 @@
...
-          divide(100, l);
-        } catch (string& e) {
-          printf("%s\n", e.c_str());
-        } catch (ZeroDivide e2) {
-          printf("Division error\n");
+    \begin{cppcode*}{fontsize=\scriptsize,gobble=2}
+      process_stream_data(stream &s) {
+        ...
+        if (data_location >= buffer.length()) {
+          raise range_error("buf overflow");
```
bernhardmgruber, line 32:
```suggestion
          throw range_error("buf overflow");
```

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,172 @@
...
+        \item define your own subclass of \texttt{std::exception} if needed
+      \end{itemize}
+      \item exceptions will be caught if the object type in the \textit{catch} clause matches
       \begin{itemize}
-      \item instead, the {\it unexpected} handler is called
-      \item by default, it terminates the program
+          \item if nothing catches an exception then \texttt{std::terminate} is called
+          \item throw exceptions by value, catch them by (const) reference
       \end{itemize}
+      \item all objects on the stack between the \textit{throw} and the \textit{catch} are destructed automatically
```
bernhardmgruber, line 53: Should we mention that this is called `stack unwinding`?

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,172 @@
...
+      \begin{itemize}
+        \item this should give you a clean release of intermediate resources
+        \item make sure you are using the RAII idiom in your own classes
+      \end{itemize}
+      \item use exceptions for \textit{unlikely} runtime errors outside the program's control (bad inputs, files unexpectedly not found, DB connection, \ldots)
+      \item \textit{don't} use exceptions for logic errors in your code (consider \texttt{assert} and tests)
```
bernhardmgruber, line 59: Maybe add:
don't use exceptions to provide no or an alternative return value. Use `std::optional`, `std::variant` or `boost::outcome` (or `std::expected` in C++23)`

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,172 @@
...
+    \begin{cppcode*}{fontsize=\tiny,gobble=2}
+      // For this code logic C-style error codes
+      //   may be better
+      for (const &item: list) {
+        auto err_code = check_item(item);
+        if (!err_code) {
+          ... // process item
+        } else {
+          ... // warn and carry on
+        }
```
bernhardmgruber, line 148: I think this is a good example for `std::optional<T>` if you don't care about the error code. Otherwise `boost::result<T>` or `std::expected`.

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,172 @@
...
+
 \begin{frame}[fragile]
-  \frametitlecpp[11]{Full power of noexcept}
-  \begin{block}{3 uses of noexcept}
+  \frametitlecpp[11]{noexcept}
+  \begin{block}{noexcept}
     \begin{itemize}
-    \item standalone
-      \begin{cppcode*}{gobble=2, linenos=false}
+      \item specifying the \textit{noexcept} specifier states that a function is guaranteed throw an exception
```
bernhardmgruber, line 159:
```suggestion
      \item specifying the \textit{noexcept} specifier states that a function is guaranteed to not throw an exception
```

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,172 @@
...
+      \begin{itemize}
+        \item either no exceptions will be thrown or they are handled internally
+        \item checked at compile time, so it allows the compiler to optimise around that knowledge
+      \end{itemize}
+      \item \textit{noexcept(expression)} specifier evaluates the expression and, if \textit{true}, the function won't throw
+        \begin{cppcode*}{fontsize=\scriptsize,gobble=2,linenos=false}
+          int safe_if_long_is_8_bytes() noexcept(sizeof(long)==8);
+        \end{cppcode*}
+      \item Use \textit{noexcept} on leaf functions where you can be sure of the behaviour
+      \item Note that since C++11 destructors are \textit{noexcept} - never try and throw from them
```
bernhardmgruber, line 172:
```suggestion
      \item Note that since C++11 destructors are \textit{noexcept} - never throw from them
```
I missinterpreted your sentence as: I should not use a `try` block in a dtor, which you can use.

talk/morelanguage/exceptions.tex
bernhardmgruber: @graeme-a-stewart Maybe you missed this comment.

talk/morelanguage/exceptions.tex
theanalyst: Is it actually ok to suggest returning a custom struct/pair of outval,error_code as an alternative as well (and also mention it is not preferred to manipulate global errno like C)

talk/morelanguage/exceptions.tex
graeme-a-stewart: Fair point - now I say more explicitly "error code or object". I am a bit reluctant to get into the details of all the options in this slide (after all, it's the "exception" to exceptions, 😄).
- `boost::outcome` is out of scope for a core C++ course (Boost isn't really mentioned at all)
- `std::expected` didn't make it to C++20 AFAICS
- `std::system_error` is a lot heavier than an `int` (this is a tight loop, by definition)

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,172 @@
...
-  \frametitlecpp[98]{Exceptions}
-  \begin{block}{Rules}
-    \begin{itemize}
-    \item exception will skip all code until next {\it catch}
-    \begin{itemize}
-      \item still destructors are called when exiting scopes
-      \item but your own cleanup may not be
-    \end{itemize}
-    \item {\it catch} is selective on the exception type
+      \item add an exception handling block with {\it try ... catch} when exceptions are possible
```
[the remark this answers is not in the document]
graeme-a-stewart: It's a good point - at the risk of being wordy I added "...and can be handled"

talk/morelanguage/exceptions.tex
graeme-a-stewart: I split this slide in two now to make it more complete on these points (it's a bit wordy, but a lot more informative)

graeme-a-stewart: It would seem that `std::optional` and `std::variant` are better than the older `std::pair` formalism, right?

talk/morelanguage/exceptions.tex
hageboeck: With `consider assert and tests`, you went in the right direction. I wanted to prevent people from writing `assert(false)`, which I have seen too often, but that's off the table now! 👍

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,191 @@
...
+        try {
+          check_item(item); // Can throw bad_item
+          ... // process item
+        }
+        catch (bad_item &e) {
+          ... // ignore this item, but keep going
+        }
+      }
+    \end{cppcode*}
+    \columnbreak
```
hageboeck, line 157: I would like to see this in an `\begin{alertblock}{Don't}` (the red box) for discouraged things.

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,191 @@
...
+      //   may be better
+      for (const &item: list) {
+        std::optional<T> err_code = check_item(item);
+        if (!err_code) {
+          ... // process item
+        } else {
+          ... // warn and carry on
+        }
+      }
+    \end{cppcode*}
```
hageboeck, line 169: and this would go into `\begin{exampleblock}{Do}` for recommended things.

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,191 @@
...
     \begin{itemize}
-    \item throw specification is runtime only
+      \item exceptions have little cost if no exception is thrown
       \begin{itemize}
-      \item does not allow compiler optimizations
-      \item on the contrary forces extra checks
-      \item generally terminates your program if violated
+        \item a throw itself typically costs about as much as a function call (the stack needs to be unwound, but needs to be done for other error handling styles too)
+        \item so minimal if you are using exceptions for \textit{exceptional} events
+        \item thus their use is generally recommended
```
hageboeck, line 135: I would like to shorten this to create space for my next two comments below:
```suggestion
        \item so they are recommended for \textit{exceptional} events
```

hageboeck (approved):
I'm happy with this. I would change only one more thing.

graeme-a-stewart:
Thanks @hageboeck - should all be done now as per your suggestions. Using example/alert blocks in multicolumn modes is annoying, but I found a way to do it with minipages.

hageboeck (approved):
> Thanks @hageboeck - should all be done now as per your suggestions. Using example/alert blocks in multicolumn modes is annoying, but I found a way to do it with minipages.

Ouch, I didn't know that I'm sending you down a rabbit hole. 👍

talk/morelanguage/exceptions.tex
```diff
@@ -4,185 +4,203 @@
...
       \end{cppcode*}
-    \item as an expression saying whether exceptions can be sent
-      \begin{cppcode*}{gobble=2, linenos=false}
-        int f() noexcept(sizeof(long) == 8);
-      \end{cppcode*}
-    \item as an operator to know whether a function launches exceptions
-      \begin{cppcode*}{gobble=2, linenos=false}
-        constexpr bool callCannotThrow = noexcept(f());
-        if constexpr (callCannotThrow) { ... }
-      \end{cppcode*}
```
sponce, line 180: I believe this usage of noexcept returning a bool has been lost in the refactoring.

graeme-a-stewart: Yep, it did get lost - I have added it back now

Issue #62, opened 2021-11-17
Extend part about locks to shared_lock, condition_variable etc for advanced course.


hageboeck:

hageboeck:
Pretty much done with this, only question that remains:
Should we go into memory order with atomics?

bernhardmgruber:
> Pretty much done with this, only question that remains: Should we go into memory order with atomics?

I would not. Playing with memory order requires deep understanding of the C++ abstract machine (happens before, synchronizes with relationships, etc.) which well beyond the scope of a beginners or advanced course.

PR #83, opened 2022-01-17
Extend concurrency section

hageboeck:
Extend the concurrency section with:
- More info about mutex and lock (also `shared_mutex`)
- Create an extra section on `condition_variable`
  - Exercise for `condition_variable`
- Add info about atomics
   - **Note**: I didn't add anything about memory ordering. This might be opening pandoras box. The defaults in the standard are the safest and the slowest, so that should do for the students.
  - Add trivial exercise on atomics

See #62


talk/concurrency/atomic.tex
```diff
@@ -0,0 +1,95 @@
...
+      \begin{umlseqdiag}
+        \umlobject[x=0]{Thread 1}
+        \umlobject[x=3, fill=blue!20]{atomic}
+        \umlobject[x=6]{Thread 2}
+        \begin{umlcall}[op=fetch\_add]{Thread 1}{atomic}
+        \end{umlcall}
+        \begin{umlcall}[op=fetch\_add, dt=9]{Thread 2}{atomic}
+        \end{umlcall}
+      \end{umlseqdiag}
+      \draw[-triangle 60](9,0) -- (9,-4) node[right, pos=0.5]{time};
```
sponce, line 81: Arrow is too long.
If you shorten it, I would use the new space to mention that `+=` is another name for fetch_add and other operators are also thread safe. And maybe repeat that ` a+= 2` is not at all equivalent to `a = a + 2` in this case.

sponce:
Very nice. Much better than my previous slides on condition variables in particular

talk/concurrency/atomic.tex
hageboeck: Good point, done!

PR #88, opened 2022-01-26
Create a "basic" version of this course

sponce:
by default the full version is created, but setting boolean "onlybasics" to "true" in the main C++Course.tex file changes this.

### DIFF
--- a/code/ExercisesCheatSheet.md
+++ b/code/ExercisesCheatSheet.md
@@ -57,15 +57,19 @@ Day 3
 
 Exercises of day 3 are special as they will have been played (quickly) during the course. But people should replay them and discover the tools by themselves.
 
-### Compiler / `polymorphism`
+### Compiler
 
-This reuses the polymorphism exercise.
+Based on the functions exercise, we are replaying the compilation steps manually.
 
 A "break the ice" exercise here would be to ask people to do the first step (preprocessor) and make a poll on how many lines of C++ are present in the output. It will vary depending on setups, versions, OS, etc...
 
 Then the important part is to play with nm and the -C option of C++filt to be able to decode symbols, find them and thus address a build error that mentions missing symbols.
 
 Another important bit is ldd and inspecting library dependencies.
+
+Notes:
+- `--std=` >= c++11 is needed to initialise the structs.
+- The Makefile skips the generation of the object file for functions.cpp. This might surprise some people.
 
 ### `debug`
 
--- a/talk/C++Course.tex
+++ b/talk/C++Course.tex
@@ -6,6 +6,9 @@
 
 \input{setup}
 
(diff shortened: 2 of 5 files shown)


hageboeck:
I had a first look at the basic course.

This is what I noticed:
- [ ] Slide 83. Exercise regarding the compiler chain, but we are going to `code/polymorphism` to try it out. Maybe it's better to take a different example, since OO wasn't discussed yet?
- [ ] Exercise `randomise` (a.k.a. debugging) uses heap allocation of an array, but that's first (?) discussed in OO.
- [ ] 210 we're `move`ing `unique_ptr` without having explained what `move` is

sponce:
I see that the move of chapters around is creating a couple of issues indeed. Not sure how to handle it in general. In the particular cases you raise :
  - for the compiler chain exercise, indeed, pretty much any code would do, so we could (and thus should) reuse another exercise
  - for the debug exercise, I'm not sure that missing the concept of heap is too problematic for the exercise. The essential point is to find out the bug which is somehow a syntax error, leading to a bad size
  - for the unique_ptr, it's complicated. We are presenting unique_ptr without the move semantic. There is a certain level of contradiction here, so we cannot avoid really the problem. This only points out that modern C++ without move is a no go. Not sure what to decide here.

hageboeck:
> I see that the move of chapters around is creating a couple of issues indeed. Not sure how to handle it in general. In the particular cases you raise :
> 
> * for the compiler chain exercise, indeed, pretty much any code would do, so we could (and thus should) reuse another exercise

OK, maybe let's just go to the functions exercise, then. It's the first after `hello`, and it includes a header, which you would see when the preprocessor is done with it. I just pushed two commits that are doing just that.

> * for the debug exercise, I'm not sure that missing the concept of heap is too problematic for the exercise. The essential point is to find out the bug which is somehow a syntax error, leading to a bad size

That's probably right. We can leave it to the tutors, I guess.

> * for the unique_ptr, it's complicated. We are presenting unique_ptr without the move semantic. There is a certain level of contradiction here, so we cannot avoid really the problem. This only points out that modern C++ without move is a no go. Not sure what to decide here.

In the end, we might get away with explaining `std::move` as "I am willing to give up ownership now". Should we add a slide?

sponce:
So we are left with the issue of `unique_ptr` being presented without move semantic. Reading back slide 210, I see that below bullet "has only move semantic", we have a subbullet "the pointer is only owned once", which basically covers more or less the basic course by stating what it means.  The example code even has a comment "transfer of ownership" on the `std::move` call. Providing we skip the quizz just after (maybe I should make it disappear in the basic course ?), I think it's kind of fine at the end.

What do you think @hageboeck ? If you're happy like this, I merge this

Issue #86, opened 2022-01-19
New exercise for second day on operators


roiser:

roiser:
e.g. +, << and ()

bernhardmgruber:
A classic example we did at university was to build a class representing a rational number. It holds two integers for the numerator and denominator. The exercise is to implement the arithmetic and comparison operators.

PR #94, opened 2022-02-24
Add a new exercise on operators

bernhardmgruber:
Fixes: #86

- [x] Add a slide referring to the exercise

### DIFF
--- a/code/operators/README.md
+++ b/code/operators/README.md
@@ -1,0 +1,5 @@
+
+## Instructions
+
+* inspect main and complete the implementation of class Fraction step by step
+* you can comment out parts of main to test in between
--- a/code/operators/solution/operators.sol.cpp
+++ b/code/operators/solution/operators.sol.cpp
@@ -1,0 +1,199 @@
+#include <cassert>
+#include <iomanip>
+#include <iostream>
+#include <numeric>
+
+class Fraction {
+ public:
+  explicit Fraction(int i) : m_num(i), m_denom(1) {}
+  Fraction(int num, int denom) : m_num(num), m_denom(denom) {}
+
+  int num() const { return m_num; }
+  int denom() const { return m_denom; }
+
+  void normalize() {
+    const int gcd = std::gcd(m_num, m_denom);
+    m_num /= gcd;
+    m_denom /= gcd;
+  }
+
+  Fraction& operator*=(int i) {
+    m_num *= i;
+    normalize();
+    return *this;
+  }
+
+  Fraction& operator*=(Fraction f) {
+    m_num *= f.num();
+    m_denom *= f.denom();
+    normalize();
+    return *this;
+  }
+
+  Fraction& operator+=(int i) {
+    m_num += i * m_denom;
+    return *this;
+  }
+
+  Fraction& operator+=(Fraction f) {
+    m_num *= f.denom();
+    m_num += f.num() * m_denom;
+    m_denom *= f.denom();
+    normalize();
+    return *this;
+  }
+
+ private:
+  int m_num, m_denom;
+};
+
+std::ostream& operator<<(std::ostream& os, Fraction r) {
+  os << r.num() << "/" << r.denom();
+  return os;
+}
+
+Fraction operator*(Fraction r, int i) { return r *= i; }
+
+Fraction operator*(int i, Fraction r) { return r * i; }
+
+Fraction operator*(Fraction a, Fraction b) { return a *= b; }
+
+Fraction operator+(Fraction r, int i) { return r += i; }
+
+Fraction operator+(int i, Fraction r) { return r + i; }
+
+Fraction operator+(Fraction a, Fraction b) { return a += b; }
+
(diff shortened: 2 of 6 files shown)


hageboeck (changes requested):
Hello,

I added two suggestions.

The question I want to ask is this:
Should we write this more as a test, i.e. it fails and/or returns a non-zero exit code?
cout is good, but it's way too much to read through. I could imagine something like `assert()`, even though assert isn't the actual thing we want.

code/operators/README.md
```diff
@@ -0,0 +1,5 @@
+
+## Instructions
+
+* look at the code (it’s the solution to the stl exercise)
+* use lambdas to simplify it
```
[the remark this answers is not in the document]
bernhardmgruber: Thx, done!

bernhardmgruber:
> Should we write this more as a test, i.e. it fails and/or returns a non-zero exit code?
> cout is good, but it's way too much to read through. I could imagine something like assert(), even though assert isn't the actual thing we want.

I briefly thought that `main` should actually be a unit test and that it might be awesome to show the students how to use a unit test framework and write tests. But obviously, this goes *far* beyond what they know and are supposed to focus on at this point in the course. But you are right, we could just add a small `ASSERT` macro ourselves like typically found in unit tests.

hageboeck:
> > Should we write this more as a test, i.e. it fails and/or returns a non-zero exit code?
> > cout is good, but it's way too much to read through. I could imagine something like assert(), even though assert isn't the actual thing we want.
> 
> I briefly thought that `main` should actually be a unit test and that it might be awesome to show the students how to use a unit test framework and write tests. But obviously, this goes _far_ beyond what they know and are supposed to focus on at this point in the course. But you are right, we could just add a small `ASSERT` macro ourselves like typically found in unit tests.

I was playing with the exercise myself. I'll add `bool printAndTest(const std::string & what, const Fraction & result, const Fraction & target)`.
Any suggestions for the name and the name of the arguments?

hageboeck:
> I was playing with the exercise myself. I'll add `bool printAndTest(const std::string & what, const Fraction & result, const Fraction & target)`.
> Any suggestions for the function name and the name of the arguments?

It's here: https://github.com/bernhardmgruber/cpluspluscourse/pull/1

code/operators/solution/operators.sol.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+  void normalize() {
+    const int gcd = std::gcd(m_num, m_denom);
+    m_num /= gcd;
+    m_denom /= gcd;
+  }
+
+  Fraction& operator*=(int i) {
+    m_num *= i;
+    m_denom *= i;
```
sponce, line 22:
```suggestion
```

code/operators/solution/operators.sol.cpp
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  Fraction& operator+=(int i) {
+    m_num += i * m_denom;
+    return *this;
+  }
+
+  Fraction& operator+=(Fraction f) {
+    const int myDenom = m_denom;
+    *this *= f.denom();
```
sponce, line 40: that's calling the operator I've changed above. Indeed the code works, but I would change it to avoid having a public operator*(int) which actually does not multiply the fraction by the int, but rather "denormalize" it

code/operators/solution/operators.sol.cpp
bernhardmgruber: I am embarassed. Thx for spotting that!

code/operators/solution/operators.sol.cpp
bernhardmgruber: Sure, I will adapt it accordingly!

code/operators/solution/operators.sol.cpp
sponce: No problem. That's the whole purpose of reviewing code !

code/operators/solution/operators.sol.cpp
```diff
@@ -0,0 +1,196 @@
...
+  int denom() const { return m_denom; }
+
+  void normalize() {
+    const int gcd = std::gcd(m_num, m_denom);
+    m_num /= gcd;
+    m_denom /= gcd;
+  }
+
+  Fraction& operator*=(int i) {
+    m_num *= i;
```
sponce, line 21:
```suggestion
    m_num *= i;
    normalize();
```

bernhardmgruber: Thx! I also added a new `cout` to check this case as well.

sponce:
@hageboeck is it ok if I merge this as is and you rebase https://github.com/bernhardmgruber/cpluspluscourse/pull/1 and create a new pull request ?

hageboeck:
> @hageboeck is it ok if I merge this as is and you rebase [bernhardmgruber#1](https://github.com/bernhardmgruber/cpluspluscourse/pull/1) and create a new pull request ?

@sponce OK, go ahead!

PR #95, opened 2022-02-24
Add a subsection on profiling with perf/vTune

bernhardmgruber:
Fixes: #67

### DIFF
--- a/talk/tools/profiling.tex
+++ b/talk/tools/profiling.tex
@@ -1,0 +1,42 @@
+\subsection[prof]{Profiling}
+
+\begin{frame}[fragile]
+  \frametitle{Profiling}
+  \begin{block}{Conceptually}
+    \begin{itemize}
+      \item take a measurement of a performance aspect of a program
+      \begin{itemize}
+        \item where in my code is most of the time spent?
+        \item is my program compute or memory bound?
+        \item does my program make good use of the cache?
+        \item is my program using all cores most of the time?
+        \item how often are threads blocked and why?
+        \item which API calls are made and in which order?
+        \item ...
+      \end{itemize}
+      \item the goal is to find performance bottlenecks
+      \item is usually done on a compiled program, not on source code
+    \end{itemize}
+  \end{block}
+\end{frame}
+
+\begin{frame}[fragile]
+  \frametitle{perf, VTune and uProf}
+  \begin{block}{perf}
+    \begin{itemize}
+      \item perf is a powerful command line profiling tool for linux
+      \item compile with \mintinline{bash}{-g -fno-omit-frame-pointer}
+      \item \mintinline{bash}{perf stat -d <prg>} gathers performance statistics while running \mintinline{bash}{<prg>}
+      \item \mintinline{bash}{perf record -g <prg>} starts profiling \mintinline{bash}{<prg>}
+      \item \mintinline{bash}{perf report} displays a report from the last profile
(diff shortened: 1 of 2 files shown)


bernhardmgruber:
@amadio I added a link to your fabulous talk on perf!

talk/tools/profiling.tex
```diff
@@ -0,0 +1,42 @@
...
+    \item \mintinline{bash}{perf stat -d <prg>} gathers performance statistics while running \mintinline{bash}{<prg>}
+    \item \mintinline{bash}{perf record -g <prg>} starts profiling \mintinline{bash}{<prg>}
+    \item \mintinline{bash}{perf report} displays a report from the last profile
+    \item More information in {\color{blue!50!white} \href{https://perf.wiki.kernel.org/index.php/Main_Page}{this wiki}} or in {\color{blue!50!white} \href{https://indico.cern.ch/event/980497/contributions/4130271/attachments/2161581/3647235/linux-systems-performance.pdf}{this talk}}
+  \end{itemize}
+\end{block}
+\begin{block}{VTune}
+  \begin{itemize}
+    \item Graphical profiler from Intel with rich features
+    \item Needs Intel CPU for full experience
```
hageboeck, line 38:
```suggestion
    \item Higher level of detail on Intel CPU, but others work, too
```

hageboeck (approved):
LGTM. I tried to weaken a bit the statement about what CPU you need for profiling.

talk/tools/profiling.tex
amadio: Have you tried VTune on AMD CPU?

amadio: Maybe add a link to this for AMD: https://developer.amd.com/amd-uprof/

hageboeck: I was only listening to what Intel people said. Not tried myself.
According to some forum posts it might be even worse than they said, so I'll take that back.

bernhardmgruber: I have an AMD Ryzen and I tried vTune a few months back and it only showed an error when I wanted to start a profiling session. I did not try more than 5min though, and only with the general hotspot analysis I think.

bernhardmgruber: @amadio did you ever try amd-uprof? I never did. I would prefer only recommending tools that we know work and are useful.

amadio: Yes, I tried it, it looks alright. Not as powerful as VTune, but not bad.

talk/tools/profiling.tex
```diff
@@ -0,0 +1,42 @@
...
+\begin{frame}[fragile]
+  \frametitle{perf, VTune and uProf}
+  \begin{block}{perf}
+    \begin{itemize}
+      \item perf is a powerful command line profiling tool for linux
+      \item compile with \mintinline{bash}{-g -fno-omit-frame-pointer}
+      \item \mintinline{bash}{perf stat -d <prg>} gathers performance statistics while running \mintinline{bash}{<prg>}
+      \item \mintinline{bash}{perf record -g <prg>} starts profiling \mintinline{bash}{<prg>}
+      \item \mintinline{bash}{perf report} displays a report from the last profile
+      \item More information in {\color{blue!50!white} \href{https://perf.wiki.kernel.org/index.php/Main_Page}{this wiki}} or in {\color{blue!50!white} \href{https://indico.cern.ch/event/980497/contributions/4130271/attachments/2161581/3647235/linux-systems-performance.pdf}{this talk}}
```
amadio, line 32: Why do you set link color "by hand"? You know that it can be set globally in the hyperref package's settings, right?

For `perf` I also strongly recommend adding a link to https://www.brendangregg.com/linuxperf.html.

bernhardmgruber: The link color was now set by @hageboeck upstream, so the manual coloring is no longer required. I will remove it.

PR #96, opened 2022-02-24
Small improvements to the basic chapter

bernhardmgruber:
I read through the basic chapter and here are some suggested changes.

### DIFF
--- a/talk/basicconcepts/classenum.tex
+++ b/talk/basicconcepts/classenum.tex
@@ -1,4 +1,4 @@
-\subsection[Compound]{Compound data types}
+\subsection[Class/Enum]{Class and enum types}
 
 \begin{frame}[fragile]
   \frametitlecpp[98]{struct}
@@ -125,12 +125,12 @@
     \begin{cppcode*}{}
       enum VType { Bus, Car }; enum Color { Red, Blue };
       VType t = Bus;
-      if (t == Red) {  // We do enter !  }
+      if (t == Red) { /* We do enter */ }
       int a = 5 * Car; // Ok, a = 5
 
       enum class VT { Bus, Car }; enum class Col { Red, Blue };
       VT t = VT::Bus;
-      if (t == Col::Red) { // Compiler error }
+      if (t == Col::Red) { /* Compiler error */ }
       int a = t * 5;       // Compiler error
     \end{cppcode*}
   \end{exampleblock}
--- a/talk/basicconcepts/control.tex
+++ b/talk/basicconcepts/control.tex
@@ -1,15 +1,15 @@
-\subsection[Control]{Control instructions}
+\subsection[Control]{Control structures}
 
 \begin{frame}[fragile]
-  \frametitlecpp[98]{Control instructions : if}
+  \frametitlecpp[98]{Control structures : if}
   \begin{block}{if syntax}
     \begin{cppcode*}{}
       if (condition1) {
-        Instructions1;
+        Statements1;
       } else if (condition2) {
-        Instructions2;
+        Statements2;
       } else {
-        Instructions3;
+        Statements3;
       }
     \end{cppcode*}
     \begin{itemize}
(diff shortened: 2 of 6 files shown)


talk/basicconcepts/classenum.tex
```diff
@@ -1,4 +1,4 @@
-\subsection[Compound]{Compound data types}
+\subsection[Class/Enum]{Class and enum types}
```
bernhardmgruber, line 1: The term compound type is used wrongly here, because compound types includes e.g. references and pointers. What is meant here are class and enum types. Unions, weirdly, are class types.

talk/basicconcepts/control.tex
```diff
@@ -1,15 +1,15 @@
-\subsection[Control]{Control instructions}
+\subsection[Control]{Control structures}
```
bernhardmgruber, line 1: I disliked the term control instructions, since they are not instructions but impose structures on the code. We could also call them `Control statements`, but control structures is probably the most used term.

talk/basicconcepts/functions.tex
```diff
@@ -248,7 +248,7 @@
             // Step 2: muons
             for (....) {
               if (...) {
-                data.delete(...);
+                data.erase(...);
```
bernhardmgruber, line 251: `delete` was syntax highlighted here, but I guess a member function was meant here and not `operator delete`.

talk/basicconcepts/scopesnamespaces.tex
```diff
@@ -8,7 +8,7 @@
     \begin{itemize}
     \item simple block of code, within \mintinline{cpp}{{}}
     \item function, class, namespace
-    \item translation unit for global declarations
+    \item translation unit (.cpp file + all includes) for global declarations
```
bernhardmgruber, line 11: I figured the term translation unit needed a short explanation.

talk/basicconcepts/scopesnamespaces.tex
```diff
@@ -31,12 +31,12 @@
...
+  \begin{block}{Variable life time}
     \begin{itemize}
-      \item Resources are allocated when declared
-      \item Resources are freed at the end of a scope
-      \item Good practice: initialise resources when allocating!
+      \item Variables are allocated when defined
+      \item Variables are freed at the end of a scope
+      \item Good practice: initialise variables when allocating them!
     \end{itemize}
   \end{block}
```
bernhardmgruber, line 41: That slide confused me initially. It is the first time we talk about a resource without defining what it is. Then I thought we wanted to talk about RAII. But that only really makes sense after we know about constructors/destructors. So I concluded this slide talks about when variables are allocated and destroyed. So I chose to replace resource by variable. Technically, the term `object` would be more appropriate, but we have a big OOP block later, where the term `object` is then used in the OOP sense.

talk/basicconcepts/scopesnamespaces.tex
```diff
@@ -103,24 +103,25 @@
...
           p::inner::a = 4;
           inner::a = 4;
           n::a = 5;
         }
       }
       using namespace p::inner;
-      int g() {
-        a = 3; // using p::inner
+      void g() {
+        a = 3; // error: ambiguous
```
bernhardmgruber, line 124: I don't know if this is the best way to present this example, but this line is definitely a compile error. Maybe it's also a good idea to present it, so people know that we can get into ambiguities with if we pull several namespaces into each other.

talk/basicconcepts/scopesnamespaces.tex
sponce: You're absolutely right. And on top I propose to move this part earlier in the talk in #98 so we should stick to simple concepts like variables here.

talk/basicconcepts/scopesnamespaces.tex
sponce: At least, the slide was buggy and this fixes it. I kind of like the fact that we immediately show the drawbacks of using `using` too much.

PR #98, opened 2022-02-28
Move namespace presentation before compound types, as they are used t…

sponce:


sponce:
@hageboeck could you double check this and merge if it's ok ?

PR #103, opened 2022-03-08
Small improvements to the modern C++ chapter

bernhardmgruber:
WIP


talk/morelanguage/constexpr.tex
```diff
@@ -75,20 +75,20 @@
 \begin{frame}[fragile]
   \frametitlecpp[11]{Real life example}
   \begin{cppcode*}{}
-    constexpr float toSI(const float v, const char unit) {
+    constexpr float toSI(float v, char unit) {
```
bernhardmgruber, line 78: Do we have a style recommendation for making by-value parameters `const`? It's not a bad thing per se, but adds some noise to the slides.

talk/morelanguage/lambda.tex
```diff
@@ -169,20 +169,20 @@
...
   \pause
-  \begin{block}{by reference}
+  \begin{block}{all by reference}
     \begin{cppcode*}{linenos=false}
       [&](...) { ... };
     \end{cppcode*}
   \end{block}
   \pause
-  \begin{block}{exceptions}
+  \begin{block}{mix}
```
bernhardmgruber, line 185: The work "exceptions" is confusing here.

talk/morelanguage/move.tex
```diff
@@ -11,8 +11,9 @@
         b = c;
       }
       std::vector<int> v(10000), w(10000);
-      for (int i = 0; i < 10000; i++) v[i] = i;
-      for (int i = 0; i < 10000; i++) w[i] = i*i;
+      for (int i = 0; i < 10000; i++) {
+        v[i] = i; w[i] = i*i;
+      }
```
bernhardmgruber, line 16: I found that easier to read, but you are free to debate me.

talk/morelanguage/move.tex
```diff
@@ -49,14 +51,14 @@
   \frametitlecpp[11]{Move semantics : the problem}
   \begin{exampleblock}{Another potentially non efficient code}
     \begin{cppcode*}{}
-      std::vector<int> vrandom(unsigned int n) {
-        std::vector<int> result(n);
+      MyVector<int> vrandom(unsigned int n) {
+        MyVector<int> result(n);
```
bernhardmgruber, line 55: `std::vector` will move nicely here, so I chose a custom type.

talk/morelanguage/move.tex
```diff
@@ -93,12 +95,12 @@
...
     \begin{itemize}
-      \item a new type of reference : rvalue references
+      \item a new type of reference : rvalue reference
       \begin{itemize}
       \item used for move semantic
-      \item denoted by \&\&
+      \item denoted by \mintinline{cpp}{&&}
       \end{itemize}
-      \item 2 new members in every class, with move semantic :
+      \item 2 new members in every class, with move semantic:
```
bernhardmgruber, line 103: I have seen this in many places in the presentation. Is there a reason the colon is not attached to the last word?

talk/morelanguage/move.tex
```diff
@@ -120,16 +122,16 @@
   \frametitlecpp[11]{Move semantics}
   \begin{block}{A few important points concerning move semantic}
     \begin{itemize}
-    \item the whole STL understands move semantics
-    \item move assignment operator is allowed to destroy source
+    \item the language and STL understand move semantics
+    \item move constructor and assignment operator are allowed to leave source "empty"
```
bernhardmgruber, line 126: I prefer the wording that the moved from object is left "empty" than destroyed, because the moved from object is in fact not destroyed, as no destructor is run.

talk/morelanguage/move.tex
```diff
@@ -156,16 +158,16 @@
...
     \end{itemize}
     \begin{cppcode*}{firstnumber=6}
-      T a;
-      T b = a;                   // Copy assign
-      T c = static_cast<T&&>(a); // Move assign
-      T d = std::move(a);        // Move assign
+      T a, b;
+      b = a;                   // Copy assign
+      b = std::move(a);        // Move assign
+      b = static_cast<T&&>(a); // Move assign (don't)
```
bernhardmgruber, line 170: Funnily enough, `T b = a` is a copy construction, because the compiler rewrites this line to `T b(a)`;

talk/morelanguage/move.tex
```diff
@@ -205,24 +208,47 @@
...
+        Movable& operator=(Movable&& other) noexcept {
+          swap(*this, other);
+          return *this;
+        }
+        friend void swap(Movable &a, Movable &b) noexcept { ... }
+      };
+    \end{cppcode*}
+  \end{exampleblock}
+\end{frame}
+
```
bernhardmgruber, line 251: I also felt we should show the version without the unified copy/move assignment operator.

talk/morelanguage/stl.tex
```diff
@@ -58,7 +58,7 @@
     \item examples
       \begin{itemize}
       \item string
-      \item tuple, list, vector, deque
+      \item tuple, list, forward\_list, vector, deque
```
bernhardmgruber, line 61: Forward list is C++11, should we change the slide header? I wanted to include it here though, since we give an overview.

talk/morelanguage/stl.tex
```diff
@@ -97,27 +97,28 @@
     \item irrespective of
       \begin{itemize}
       \item the data handled
-      \item the container where data live
-      \item the iterator used to go through data
+      \item the container where the data live
+      \item the iterator used to go through data (almost)
```
bernhardmgruber, line 101: Algorithms are usually sensitive to the iterator category.

talk/morelanguage/stl.tex
```diff
@@ -97,27 +97,28 @@
...
       \begin{itemize}
       \item for\_each, find, find\_if, count, count\_if, search
       \item copy, swap, transform, replace, fill, generate
       \item remove, remove\_if
-      \item unique, reverse, rotate, random, partition
-      \item sort, partial\_sort, merge, min, max
-      \item lexicographical\_compare, iota, accumulate, partial\_sum
+      \item unique, reverse, rotate, shuffle, partition
+      \item sort, partial\_sort, merge, make\_heap, min, max
+      \item lexicographical\_compare, iota, reduce, partial\_sum
```
bernhardmgruber, line 110: `random` does not exist and you should prefer `reduce` over `accumulate` if your `T` and operation are commutative.

talk/morelanguage/stl.tex
```diff
@@ -163,22 +165,40 @@
...
+    list<int>::iterator it =
+      find_if(l.begin(), l.end(),
               compose2(logical_and<bool>(),
                        bind2nd(greater_equal<int>(), 1),
                        bind2nd(less_equal<int>(), 10)));
 
     // Computes sin(x)/(x + DBL_MIN) for elements of a range.
     transform(first, last, first,
-              compose2(divides<double>(),
+              compose2(divides<double>(), // non-standard
```
bernhardmgruber, line 179: I don't know what to do here. `compose2` is not standard, but available in some STL implementations. The alternative is `std::bind`, which is C++11.

talk/morelanguage/stl.tex
```diff
@@ -163,22 +165,40 @@
...
+    ...
+    const auto it =
+      find_if(l.begin(), l.end(),
+        [](int i) { return i >= 1 && i <= 10; });
+
+    // Computes sin(x)/(x + DBL_MIN) for elements of a range.
+    transform(first, last, first,
+      [](auto x) { return sin(x)/(x + DBL_MIN); });
+  \end{cppcode}
 \end{frame}
```
bernhardmgruber, line 202: Although we discuss lambdas later, I found it important to show how much more readable the code gets with them.

talk/morelanguage/exceptions.tex
```diff
@@ -210,7 +211,7 @@
   \frametitlecpp[11]{noexcept operator}
   \begin{block}{}
     \begin{itemize}
-      \item the operator \textit{noexcept(expression)} performs a compile time check to know whether an expression can launch exceptions
+      \item the \mintinline{cpp}{noexcept(expression)} operator performs a compile time check to know whether an expression can throw exceptions
```
hageboeck, line 214:
```suggestion
      \item the \mintinline{cpp}{noexcept(expression)} operator performs a compile-time check to know whether an expression can throw exceptions
```

talk/morelanguage/move.tex
hageboeck: And copy elision will take care of `MyVector` ...
Maybe we should think about something for the future.

talk/morelanguage/move.tex
hageboeck: That's how you typeset in French texts. I think it's an oversight when this makes it into the text.

talk/morelanguage/move.tex
```diff
@@ -120,16 +122,16 @@
...
-      \item still, I advice to never leave inconsistent objects
+      \item so do not use source afterward
+      \item leave source in a valid state (for its destructor)
       \end{itemize}
-    \item if not implemented, move falls back to copy version
+    \item if move semantic is not implemented, copy will be performed
     \item move is called by the compiler whenever possible
       \begin{itemize}
-      \item e.g. when passing temporary
+      \item e.g. when passing temporaries or returning from a function
```
hageboeck, line 134:
```suggestion
      \item e.g.\ when passing temporaries or returning from a function
```
This will fix French spacing in latex.

talk/morelanguage/move.tex
hageboeck: That's correct. Maybe change it back and label as `Copy construction`?

We don't need to mention that in the lecture, but if somebody asks, I would be happy to explain.

talk/morelanguage/move.tex
```diff
@@ -205,24 +208,47 @@
...
-      void swap(Movable &a, Movable &b);
       Movable a, b;
       a = b;            // operator= copies b into "other"
       a = std::move(b); // operator= moves b into "other"
     \end{cppcode*}
   \end{exampleblock}
 \end{frame}
 
+\begin{frame}[fragile]
+  \frametitlecpp[11]{Move semantics : the complete picture}
```
hageboeck, line 230:
```suggestion
  \frametitlecpp[11]{Move semantics: the complete picture}
```

talk/morelanguage/stl.tex
hageboeck: No opinion on that. It's good for me with and without.

hageboeck (approved):
Good work. I added two small suggestions.

talk/morelanguage/move.tex
hageboeck: I found it hard to spot the differences, as the code moves up and down. I suggest to to beamer's `\begin{overprint}` inside the green box, so the box would stay stationary, and the code would start at the identical location.

Does this make sense?

talk/morelanguage/move.tex
bernhardmgruber: Oh, I did not know. Thx! I will revert those changes then!

talk/morelanguage/move.tex
bernhardmgruber: I changed the snippet now to use the `swap` implementation on the same slide and labelled what operations are performed accordingly.

talk/morelanguage/stl.tex
bernhardmgruber: I added C++11 in parenthesis. I also noticed `std::array` is missing. And I replaced `hash_` with `unordered_`.

talk/morelanguage/exceptions.tex
bernhardmgruber: thx!

talk/morelanguage/move.tex
bernhardmgruber: thx!

talk/morelanguage/move.tex
bernhardmgruber: I will not mess with French typesetting, if it is used consistently in the presentation.

talk/morelanguage/move.tex
bernhardmgruber: I tried it but then there was more vertical space between the frame titel and the exampleblock and the code did no longer fit. I now aligned the frame content at the top, that works better.

talk/morelanguage/constexpr.tex
```diff
@@ -55,7 +55,7 @@
   \frametitlecpp[14]{Generalized Constant Expressions(2)}
    \begin{alertblock}{Few limitations in C++14 (more in C++11)}
     \begin{itemize}
-    \item function's body cannot contain try-catch or static variables
+    \item function's body cannot contain try-catch, uninitialized or static variables - details on \href{https://en.cppreference.com/w/cpp/language/constexpr}{cppreference}
```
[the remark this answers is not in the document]
bernhardmgruber: I mean because of an additional comma after `uninitialized`?

talk/morelanguage/move.tex
bernhardmgruber: Not necessarily. The C++ standard does not guarantee named return value optimization NRVO. A debug build of that program would copy, if `MyVector` does not implement a move constructor.

PR #113, opened 2022-03-11
Update exercise introduction

bernhardmgruber:
* add Stephan and me as co-authors
* update date
* add that clang tools should be installed
* update download link for exercises

### DIFF
--- a/code/exercisesIntro.tex
+++ b/code/exercisesIntro.tex
@@ -40,9 +40,9 @@
 \newcommand{\cpp}{C$^{++}$}
 
 \title{Exercises HowTo}
-\author[S. Ponce]{S\'ebastien Ponce  \texttt{sebastien.ponce@cern.ch}}
+\author[B. Gruber, S. Hageboeck, S. Ponce]{Based on the work of \\ S\'ebastien Ponce \\ \texttt{sebastien.ponce@cern.ch}}
 \institute{CERN}
-\date{August 2021}
+\date{March 2022}
 \pgfdeclareimage[height=0.5cm]{cernlogo}{CERN-logo.jpg}
 \logo{\pgfuseimage{cernlogo}}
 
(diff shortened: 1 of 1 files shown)


code/exercisesIntro.tex
```diff
@@ -90,9 +90,10 @@
     \small
     \begin{verbatim}
 
-wget http://cern.ch/sponce/C++Course/code.tgz
-tar xzf code.tgz
-cd code
+wget https://github.com/hsf-training/cpluspluscourse/archive\
+/refs/heads/master.zip
+unzip master.zip
```
bernhardmgruber, line 95: Do people have `unzip` available? Is there a better alternative?

hageboeck: People should know how to unzip stuff.

This, however, will be the full course. Let me find out if you can create stable links (== still work when we replace the PDF with a new version) on Indico.

bernhardmgruber: It's a 1.2MiB file. I don't think it harms to also download the tex files in addition to the exercises. Also there is no direct support in GitHub to download a subfolder.

With external services, it's possible: https://download-directory.github.io/?url=https%3A%2F%2Fgithub.com%2Fhsf-training%2Fcpluspluscourse%2Ftree%2Fmaster%2Fcode. Should we use that URL?

bernhardmgruber: > Let me find out if you can create stable links (== still work when we replace the PDF with a new version) on Indico.

In case you got the wrong idea: It's about downloading the C++ source files for the exercises, not the course PDF.

hageboeck: OK, that's fine!

hageboeck: I see. Yes, let them download everything.

hageboeck (approved):

PR #115, opened 2022-03-11
Github actions for essentails version

hageboeck:
- Make essentials reachable via makefile
- Add essentials to github builds

### DIFF
--- a/talk/C++Course.tex
+++ b/talk/C++Course.tex
@@ -2,12 +2,11 @@
 \usetheme{Warsaw}
 \useoutertheme{split}
 
-%\includeonly{python/python}
-
 \input{setup}
 
 \newboolean{onlybasics}
 \setboolean{onlybasics}{false}
+\IfFileExists{onlybasics.tex}{\input{onlybasics}}{}
 
 \begin{document}
 
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -3,9 +3,15 @@ on: [push,pull_request]
 jobs:
   build_latex:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        version: [ essentials, full ]
     steps:
       - name: Set up Git repository
         uses: actions/checkout@v2
+      - name: Set up essentials course
+        if: matrix.version == 'essentials'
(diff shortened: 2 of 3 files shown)


talk/C++Course.tex
```diff
@@ -2,12 +2,11 @@
 \usetheme{Warsaw}
 \useoutertheme{split}
 
-%\includeonly{python/python}
-
 \input{setup}
 
 \newboolean{onlybasics}
 \setboolean{onlybasics}{false}
+\input{onlybasics}
```
bernhardmgruber, line 9: Could you please test first, if the file exists? Something like:
```suggestion
\IfFileExists{onlybasics.tex}{\input{onlybasics}}{}
```
This way, I do not need to create the file myself if I use an IDE like `TexStudio` or an interface such as Overleaf.

sponce:
I'm not liking too much using an extra file, but I must say that I fail finding a better solution in 15mn... there is this hack : https://tex.stackexchange.com/questions/1492/passing-parameters-to-a-document (second answer, not validated one) where you run : 
```
pdflatex "\setboolean{onlybasics}{true} \input{C++Course.tex}"
```
Although it probably does not even work with setboolean.

So I'm fine with what works :-)

sponce (approved):

hageboeck:
> I'm not liking too much using an extra file, but I must say that I fail finding a better solution in 15mn... there is this hack : https://tex.stackexchange.com/questions/1492/passing-parameters-to-a-document (second answer, not validated one) where you run :
> 
> ```
> pdflatex "\setboolean{onlybasics}{true} \input{C++Course.tex}"
> ```
> 
> Although it probably does not even work with setboolean.
> 
> So I'm fine with what works :-)

@sponce, the only other ideas I had:
- Create a `C++CourseEssentials.tex`; add to repo
- Use sed to replace the boolean, but then git will tell you that you changed the file.

I'm happy to leave this undecided until Monday, maybe somebody comes up with a solution.

bernhardmgruber (approved):

klieret (approved):

sponce:
"Perfect is the enemy of good", let's merge this.

PR #116, opened 2022-03-11
Correct signedness of plain char

graeme-a-stewart:
A plain `char` type in C++ has an implementation defined sign. e.g., x86 is signed, ARM is unsigned.

See <https://en.cppreference.com/w/cpp/language/types>.

Note that also in English one says "an N bit integer" not "an N bit**s** integer. "bit" here becomes an adjective, so it doesn't matter than there are multiple of them (cf. "an integer of 8 bit**s**", where the plural is correct).

### DIFF
--- a/talk/basicconcepts/coresyntax.tex
+++ b/talk/basicconcepts/coresyntax.tex
@@ -47,32 +47,32 @@
   \begin{cppcode}
     bool b = true;            // boolean, true or false
 
-    char c = 'a';             // 8 bits ASCII char
+    char c = 'a';             // 8 bit ASCII char
     char* s = "a C string";   // array of chars ended by \0
     string t = "a C++ string";// class provided by the STL
 
-    char c = -3;              // 8 bits integer (signed or unsigned)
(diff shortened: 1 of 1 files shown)


talk/basicconcepts/coresyntax.tex
```diff
@@ -47,32 +47,32 @@
...
     bool b = true;            // boolean, true or false
 
-    char c = 'a';             // 8 bits ASCII char
+    char c = 'a';             // 8 bit ASCII char
     char* s = "a C string";   // array of chars ended by \0
     string t = "a C++ string";// class provided by the STL
 
-    char c = -3;              // 8 bits signed integer
-    unsigned char c = 4;      // 8 bits unsigned integer
+    char c = -3;              // 8 bit integer (signed or unsigned)
```
sponce, line 54:
```suggestion
    char c = -3;              // 8 bit integer (signed or not)
```
So that it fits in the slide :-)

graeme-a-stewart: Yes - fully approve! Thanks.

PR #119, opened 2022-03-11
CMake Updates, master branch (2022.03.11.)

krasznaa:
With quite a bit more effort than I originally expected, I updated/modernised the CMake configuration of all code examples.
  - Made sure that all examples would declare/use the "`solution`" target consistently;
  - Made sure that the solutions (and the main sources where appropriate) would compile warning-free with [MSVC](https://visualstudio.microsoft.com/vs/features/cplusplus/);
    * There is still one warning remaining, but that one I can just not be bothered to get rid of right now...

```
C:\Users\akraszna\HSF C++ Tutorial\cpluspluscourse\code\memcheck\solution\Polygons.sol.cpp(15,14): warning C4996: 'strdup': The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: _strdup. See online help for details. [C:\Users\akraszna\HSF C++ Tutorial\build\memcheck\memcheckPolySol.vcxproj]
```

  - Removed the "solution" for the `constness` example, as it was the same broken source file that we give to the students to fix up/study;
  - Reduced the C\+\+ standard used by the "lambdas" exercise to C\+\+11, as it is not compatible with C\+\+17 or newer, and it prints a warning about using deprecated code with C\+\+14;
    * The CMake built projects all use C\+\+17 by default, except this one. The `Makefile` configurations don't usually even require C\+\+11.
  - Made some tiny text updates in `asan/README.md`.

A few more notes:
  - Once/if this is in, I intend to teach the CI to build all the exercises on GitHub's `ubuntu-latest`, `macos-latest` and `windows-latest` executors, so we could continuously check that they would work where they should;
  - As you can see, the main project file still excludes a couple of exercises on macOS and Windows. But all of those should be possible to update to a stage where they could be used on any platform. For instance I played with the python exercise a bit today, and it **can** be made to work on Windows. But that would probably make it become a CMake-only exercise. (Unless somebody wants to write an "export header" by hand for it.)

Pinging @roiser, @sponce, @bernhardmgruber, @hageboeck.


code/lambdas/CMakeLists.txt
```diff
@@ -3,17 +3,17 @@
 cmake_minimum_required( VERSION 3.1 )
 project( lambdas LANGUAGES CXX )
 
+# This example only works warning-free in C++11 mode with MSVC and Clang.
+set( CMAKE_CXX_STANDARD 11 )
+
```
bernhardmgruber, line 8: I usually prefer `target_compile_features(lambdas_randomize PUBLIC cxx_std_11)` since I am never sure how `CMAKE_CXX_STANDARD` interacts with different projects that are part of the same build. But up to you!

code/memcheck/CMakeLists.txt
```diff
@@ -5,23 +5,27 @@ project( memcheck LANGUAGES CXX )
 
 # Set up the compilation environment.
 include( "${CMAKE_CURRENT_SOURCE_DIR}/../CompilerSettings.cmake" )
+include( "${CMAKE_CURRENT_SOURCE_DIR}/../SolutionTarget.cmake" )
 
 # Create the user's library.
-add_library( memcheckPoly Polygons.hpp Polygons.cpp )
+add_library( memcheckPoly "Polygons.hpp" "Polygons.cpp" )
 set_target_properties( memcheckPoly PROPERTIES OUTPUT_NAME "poly" )
+if( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC" )
```
bernhardmgruber, line 13: cmake supports this directly:
```suggestion
if(MSVC)
```
appears in other places too.

bernhardmgruber (approved):
Generally, looks good to me!

code/lambdas/CMakeLists.txt
krasznaa: There are arguments to be made for both approaches. When you write a library that requires a certain C\+\+ standard from its user (who uses it independently of you) to be able to interpret its public headers, then setting this compile feature can be very good.

But I found that when building for instance the ATLAS offline project, where we control the build of everything ourselves, `CMAKE_CXX_STANDARD` is just the way to go...

code/memcheck/CMakeLists.txt
krasznaa: Interestingly enough the [MSVC](https://cmake.org/cmake/help/latest/variable/MSVC.html) variable is still considered a recommended variable to use, though I suspect it's because making it deprecated would upset too many people. 🤔 

Its "GCC equivalent" on the other hand, [CMAKE_COMPILER_IS_GNUCXX](https://cmake.org/cmake/help/latest/variable/CMAKE_COMPILER_IS_GNUCXX.html), has been deprecated since a while.

Long story short, in any new code that I write I try to only use [CMAKE_CXX_COMPILER_ID](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html) to decide what to do for different compilers. As it generally makes it a lot easier to extend the code for additional compilers down the line.

krasznaa:
Note that I don't have any rights on this repo, so if you guys are happy with it, you'll have to merge it eventually yourselves. :wink:

code/memcheck/CMakeLists.txt
bernhardmgruber: Fine for me!

PR #126, opened 2022-03-14
Add a solution for functions exercise.

hageboeck:
@krasznaa could you check that I didn't mess up the CMake?

### DIFF
--- a/code/functions/CMakeLists.txt
+++ b/code/functions/CMakeLists.txt
@@ -5,14 +5,17 @@ project( functions LANGUAGES CXX )
 
 # Set up the compilation environment.
 include( "${CMAKE_CURRENT_SOURCE_DIR}/../CompilerSettings.cmake" )
+include( "${CMAKE_CURRENT_SOURCE_DIR}/../SolutionTarget.cmake" )
 
 # Create the user's executable.
 add_executable( functions "Structs.h" "Structs.cpp" "functions.cpp" )
-if( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC" )
-   target_compile_definitions( functions PRIVATE _CRT_SECURE_NO_WARNINGS )
 
 # Create the "solution executable".
-add_executable( functions.sol EXCLUDE_FROM_ALL Structs.cpp solution/functions.sol.cpp )
-if( TARGET solution )
-   add_dependencies( solution functions.sol )
+add_executable( functions.sol EXCLUDE_FROM_ALL "Structs.h" "Structs.cpp" "solution/functions.sol.cpp" )
+target_include_directories(functions.sol PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}")
+add_dependencies( solution functions.sol )
+
+if( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC" )
+   target_compile_definitions( functions PRIVATE _CRT_SECURE_NO_WARNINGS )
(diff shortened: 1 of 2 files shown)


chavid:
I proposed one in my WIP merge request. I guess I should not have done such a corase-grain MR...

code/functions/CMakeLists.txt
```diff
@@ -5,9 +5,16 @@ project( functions LANGUAGES CXX )
...
 add_executable( functions "Structs.h" "Structs.cpp" "functions.cpp" )
+
+# Create the "solution executable".
+add_executable( functions.sol EXCLUDE_FROM_ALL "Structs.h" "Structs.cpp" "solution/functions.sol.cpp" )
+target_include_directories(functions.sol PUBLIC ${CMAKE_SOURCE_DIR})
+add_dependencies( solution functions.sol )
+
 if( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC" )
-   target_compile_definitions( functions PRIVATE _CRT_SECURE_NO_WARNINGS )
+   target_compile_definitions( functions functions.sol PRIVATE _CRT_SECURE_NO_WARNINGS )
```
krasznaa, line 19: You will need to do this one 2 separate lines/calls. The [target_compile_definitions](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html) function only affects one target at a time.

code/functions/CMakeLists.txt
```diff
@@ -5,9 +5,16 @@ project( functions LANGUAGES CXX )
...
 # Set up the compilation environment.
 include( "${CMAKE_CURRENT_SOURCE_DIR}/../CompilerSettings.cmake" )
+include( "${CMAKE_CURRENT_SOURCE_DIR}/../SolutionTarget.cmake" )
 
 # Create the user's executable.
 add_executable( functions "Structs.h" "Structs.cpp" "functions.cpp" )
+
+# Create the "solution executable".
+add_executable( functions.sol EXCLUDE_FROM_ALL "Structs.h" "Structs.cpp" "solution/functions.sol.cpp" )
+target_include_directories(functions.sol PUBLIC ${CMAKE_SOURCE_DIR})
```
krasznaa, line 15: You should instead write:

```cmake
target_include_directories(functions.sol PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}")
```

  - For executables `PRIVATE` is more reasonable than `PUBLIC`. (I would not have written a comment just about that though.)
  - If you use [CMAKE_SOURCE_DIR](https://cmake.org/cmake/help/latest/variable/CMAKE_SOURCE_DIR.html) then building this "package" as part of https://github.com/hsf-training/cpluspluscourse/blob/master/code/CMakeLists.txt would not work. [CMAKE_CURRENT_SOURCE_DIR](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_SOURCE_DIR.html) is designed exactly for cases like this. :wink:
  - It's a good idea to put quotation marks around any variable that describes a path. To be able to handle paths with spaces in them correctly.

krasznaa (commented):
Other than the technical comments, I'm of course all on board with this.

bernhardmgruber:
This PR conflicts with #121, which I would merged first. @hageboeck can you please sort this out and merge both of them in the order you prefer?

code/functions/CMakeLists.txt
hageboeck: Ah, I should have checked ...

chavid:
Merge of #121 done. Please rebase.

PR #130, opened 2022-03-14
Fix addresses in memorystack.

hageboeck:
As @amadio pointed out, the stacks are showing wrong addresses when
setting word size == 1, but pushing integers on the stack.

There's still a slightly ugly thing: pointers are 32 bits in the current version. I think we can live with that, though. Only to be mentioned if somebody asks.

### DIFF
--- a/talk/basicconcepts/arrayspointers.tex
+++ b/talk/basicconcepts/arrayspointers.tex
@@ -50,7 +50,7 @@ int l = *pak;
   \columnbreak
     \onslide<2->{
       \begin{tikzpicture}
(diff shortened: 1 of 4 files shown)


bernhardmgruber:
I also saw that briefly tried to make the addresses go in increments of 4. But then the variable names in the boxes looked really bad. This seems fixed now:
image
However, `int`s usually start on addresses which are a multiple of 4. Can you fix that as well please? Thx!

hageboeck:
Fixed the offset. Will merge after CI confirms.

sponce (approved):

PR #139, opened 2022-03-16
Change values in vector of control exercise

sponce:
The idea is to have higher chance crashing/having weird values when people do the mistake of thinking that a range based loop gives indices in the local variable. Codes like :
```cpp
for (auto i : numbers) {
   auto num = numbers[i];
   ...
}
```

### DIFF
--- a/code/control/control.cpp
+++ b/code/control/control.cpp
@@ -1,14 +1,14 @@
 #include <iostream>
 
-constexpr int numbers[]{1, 2, 3, 4, 5, 6, 7, 8, 9};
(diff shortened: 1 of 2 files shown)


sponce:
Couple of notes here : 
  - I finally put a 10^6 offset so that we get most of the time a segfault when making the mentioned mistake
  - 1000 offset was not enough. Locally I was just getting a bunch of 0s and 0 sums
  - small negative numbers were the same with the extra burden that I had to add `std::abs` in the isodd function to make it work properly when no mistake is made. I decided to not go negative and use unsigned int to make isodd correct

bernhardmgruber (approved):

PR #144, opened 2022-03-16
Tweak OO chapter for the essentials version

hageboeck:
Here are the illustrations I used in the morning lecture.

I also removed the two final parts. Let's do this in the advanced course.


bernhardmgruber:
Did you kill the CI? I don't see any passed checks.

bernhardmgruber (approved):

PR #148, opened 2022-03-16
Improve debug and polymorph exercise

bernhardmgruber:
Many students had troubles in my afternoon session so I would like to propose the following fixes.

### DIFF
--- a/code/debug/randomize.cpp
+++ b/code/debug/randomize.cpp
@@ -26,7 +26,7 @@ void createAndFillVector(int** v, unsigned int len) {
 }
 
 int main() {
-    int *v;
+    int *v = nullptr;
     // create and randomize vector
     randomize(v, LEN+1);
     createAndFillVector(&v, LEN+1);
--- a/code/polymorphism/solution/trypoly.sol.cpp
+++ b/code/polymorphism/solution/trypoly.sol.cpp
@@ -6,11 +6,11 @@ int main() {
     Pentagon penta{1.0};
     std::cout << "Penta : perimeter = " << penta.computePerimeter() << "\n\n";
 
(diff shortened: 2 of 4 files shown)


code/debug/randomize.cpp
```diff
@@ -26,10 +26,10 @@ void createAndFillVector(int** v, unsigned int len) {
 }
 
 int main() {
-    int *v;
+    int *v = nullptr;
```
bernhardmgruber, line 29: I propose to initialize the pointer. Many students inspected the crash inside `swap` in the debugger and nothing looked wrong to them. If we make `v` `nullptr` initially, then the addresses inside `swap` will look more suspicious. Also the example crashes more deterministically.

code/polymorphism/solution/trypoly.sol.cpp
```diff
@@ -6,11 +6,11 @@ int main() {
     Pentagon penta{1.0};
     std::cout << "Penta : perimeter = " << penta.computePerimeter() << "\n\n";
 
-    // create an Hexagon, call its perimeter method
+    // create a Hexagon, call its perimeter method
     Hexagon hexa{1.0};
     std::cout << "Hexa : perimeter = " << hexa.computePerimeter() << "\n\n";
 
-    // create an Hexagon, call its parent's perimeter method
+    // create a Hexagon, call the perimeter method through a reference to Polygon
```
bernhardmgruber, line 13: This should steer students in the right direction and not write `Hexagon h{2.0f}; h.Polygon::computePerimeter();`.

code/debug/randomize.cpp
theanalyst: :+1:

PR #165, opened 2022-03-18
Two small fixes and a rename

bernhardmgruber:


talk/objectorientation/static.tex
```diff
@@ -14,10 +14,10 @@
...
       static std::string upper(std::string) {...}
     private:
-      static int s_nbCallsToUpper; // add `inline` in C++17
+      static int callsToUpper; // add `inline` in C++17
     };
-    int Text::s_nbCallsToUpper = 0; // required before C++17
+    int Text::callsToUpper = 0; // required before C++17
     std::string uppers = Text::upper("my text");
-    // now Text::s_nbCallsToUpper is 1
+    // now Text::callsToUpper is 1
```
bernhardmgruber, line 21: I just figured this is a simpler name.

sponce: The "s_" is a remnant of LHCb coding conventions. It shows clearly that it is static. But agreed that coding conventions are not the topic of this slide !

bernhardmgruber: I thought so, but as you said, the students should not be confronted with too many styles and coding conventions in the talk.

PR #166, opened 2022-03-18
Add a slide on integer literals

bernhardmgruber:


sponce:
That's very useful indeed ! Shall we add another one (much shorter) and floating literals ? I essentially think of remauning people that 2.0 is a double and not a float and that they should use 2.0f when they work with floats.

bernhardmgruber:
There you go: https://github.com/hsf-training/cpluspluscourse/pull/169

PR #184, opened 2022-07-20
Produce stable links for latest version of the slides

### LINKED ISSUE #171: Create a fixed URL for the latest version of the slides

roiser:
Runs a github action (worflow/slides.yaml) which will produce the two versions of the slides (essentials & full) and commit them to the "download" branch. The github action should run on every commit, pull request and on Sunday 23:55. 

The result should be stable links to the slides, if not mistaken those should be

https://github.com/hsf-training/cpluspluscourse/blob/download/talk/C%2B%2BCourse_essentials.pdf 
https://github.com/hsf-training/cpluspluscourse/blob/download/talk/C%2B%2BCourse_full.pdf

and the base directory 

https://github.com/hsf-training/cpluspluscourse/tree/download/talk

The PR also contains the ignore of xdv files in the .gitignore. 

Fixes #171

### DIFF
--- a/.github/workflows/slides.yml
+++ b/.github/workflows/slides.yml
@@ -1,0 +1,18 @@
+name: GitHub Actions Demo
+on: [push]
+jobs:
+  Explore-GitHub-Actions:
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "🎉 The job was automatically triggered by a ${{ github.event_name }} event."
+      - run: echo "🐧 This job is now running on a ${{ runner.os }} server hosted by GitHub!"
+      - run: echo "🔎 The name of your branch is ${{ github.ref }} and your repository is ${{ github.repository }}."
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - run: echo "💡 The ${{ github.repository }} repository has been cloned to the runner."
+      - run: echo "🖥️ The workflow is now ready to test your code on the runner."
+      - name: List files in the repository
+        run: |
+          ls ${{ github.workspace }}
+      - run: echo "🍏 This job's status is ${{ job.status }}."
+
--- a/.github/workflows/slides.yml
+++ b/.github/workflows/slides.yml
@@ -6,9 +6,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: clone download-slides
+      - name: clone download
         run: |
-          git clone -b download-slides https://x-access-token:${{secrets.github_token}}@github.com/${{github.repository}}.git download-slides
+          git clone -b download https://x-access-token:${{secrets.github_token}}@github.com/${{github.repository}}.git download
       - name: Compile LaTeX document
         uses: xu-cheng/latex-action@v2
         with:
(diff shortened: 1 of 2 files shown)


roiser:
PS: Many thanks to @bernhardmgruber and @hageboeck for inspiration from their previous github actions ;-)

klieret:
Having linkable latest versions of the PDF would be really cool! :) 

(Btw, to trigger builds, you can also do `git commit --allow-empty` (if there isn't a button to do this on github))

roiser:
> Having linkable latest versions of the PDF would be really cool! :)
> 
> (Btw, to trigger builds, you can also do `git commit --allow-empty` (if there isn't a button to do this on github))

It kind of works, but we discussed it yesterday and there is a possibility we blow up the repo because we commit one pdf after the other. I'm now trying to avoid this.

roiser:
I close this PR and come back to it when the current issue is solved.

roiser:
This should work now as expected. I.e. there are two branches in the repo "emptyrepo" and "download" , the github action will

- first "reset --hard" the download branch to emptyrepo
- then will run (in parallel) the builds of the two (so far) versions of the slides and commit and push them to download

this way, once a new github action will run, the previous tip of download will point to void (after the hard reset) and should be garbage collected after some time by github.com. Can somebody confirm this last point?

klieret:
https://stackoverflow.com/questions/56020314/how-frequently-do-github-run-git-gc-on-repositories  seems to be daily.

Btw, just saw that there still is an issue with github not finding the `emptyrepo` upstream (which is strange).

A different option that I see would to do

```bash
# checkout download branch still containing old files
# create new files
# now git status will show us dirty files
# We now edit the last commit to use the new files
git commit --all --amend --no-edit
# And force push it on the server
git push --force
```

klieret:
(A completely different approach would be to upload to Google drive and then create a link for that: https://github.com/marketplace/actions/google-drive-upload-git-action)

roiser:
> https://stackoverflow.com/questions/56020314/how-frequently-do-github-run-git-gc-on-repositories seems to be daily.
> 
> Btw, just saw that there still is an issue with github not finding the `emptyrepo` upstream (which is strange).
> 
> A different option that I see would be to do
> 
> ```shell
> # checkout download branch still containing old files
> # create new files
> # now git status will show us dirty files
> # We now edit the last commit to use the new files
> git commit --all --amend --no-edit
> # And force push it on the server
> git push --force
> ```

thanks the daily gc is very fine, then I hope we should not easily blow up the repo. 

the problem with the emtpyrepo branch is probablly that this action only runs within a PR for the moment? I could imagine once merged that it should be ok. I just ran it again and it looks there are some permission problems on the upstream repo (here)

roiser:
> (A completely different approach would be to upload to Google drive and then create a link for that: https://github.com/marketplace/actions/google-drive-upload-git-action)

thanks that's good to know. we discussed this yesterday and said its probably easier if we leave everything within github b/c of credentials setup. I propose to see if the current version works. If any issues this is certainly also viable.

bernhardmgruber:
> ```shell
> git commit --all --amend --no-edit
> git push --force
> ```

That sounds like a good option. Then you don't need the empty repository branch.

bernhardmgruber:
> Just realised that the github action in the PR in the upstream repo fails with
> 
> ```
> + git push origin download
>  1 file changed, 0 insertions(+), 0 deletions(-)
>  create mode 100644 talk/C++Course_essentials.pdf
> remote: Permission to hsf-training/cpluspluscourse.git denied to github-actions[bot].
> fatal: unable to access 'https://github.com/hsf-training/cpluspluscourse.git/': The requested URL returned error: 403
> Error: Process completed with exit code 128.
> ```
> 
> not sure, maybe this is because its still inside the PR and needs to be merged first?

This was exactly the problem. I pushed the changes to master and the action was able to start committing.

bernhardmgruber:
This PR has been superseded by #225.

klieret:
Since the #225 is merged, I will close this PR then.

Issue #187, opened 2022-08-02
Add a slide on capturing `this`


bernhardmgruber:
We should add a slide explaining what it means to capture `this` or `*this` in a lambda capture list.

PR #188, opened 2022-08-02
Add a slide on capturing this

bernhardmgruber:
Fixes: #187

### DIFF
--- a/talk/morelanguage/lambda.tex
+++ b/talk/morelanguage/lambda.tex
@@ -193,6 +193,33 @@
 \end{frame}
 
 \begin{frame}[fragile]
+  \frametitlecpp[11]{Capture list - this}
+  \begin{block}{}
+    Inside a (non-static) member function, we can capture \mintinline{cpp}{this}.
+  \end{block}
+  \begin{block}{}
+    \begin{cppcode*}{gobble=2}
+      [this](...) { use(*this); };
+      [&](...) { use(*this); };
+      [&, this](...) { use(*this); };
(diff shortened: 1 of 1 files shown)


talk/morelanguage/lambda.tex
```diff
@@ -192,6 +192,33 @@
...
+  \end{block}
+  \begin{block}{}
+    \begin{cppcode*}{gobble=2}
+      [this](...) { use(*this); };
+      [&](...) { use(*this); };
+      [&, this](...) { use(*this); };
+      [=](...) { use(*this); };       // deprecated in C++20
+      [=, this](...) { use(*this); }; // allowed in C++20
+    \end{cppcode*}
+    The object at \mintinline{cpp}{*this} is captured by reference.
```
hageboeck, line 208: That might be a bit confusing. This might not fit on the slide, but how about:
```suggestion
    \mintinline{cpp}{this} is a pointer. Therefore, referring to \mintinline{cpp}{*this} is like referring to the object by reference.
```

hageboeck (approved):
Generally very nice, but I'm wondering whether we should clarify the sentence a bit or just mention it orally in the lecture.

talk/morelanguage/lambda.tex
bernhardmgruber: I tried to elaborate further.

hageboeck: That's good. Merge this if it fits on the slide. 🙂

PR #190, opened 2022-08-02
Update constexpr slides

bernhardmgruber:
Add more notes and more details on the restrictions in various language standards.
Also start with the C++14 example instead of the C++11 one.

### DIFF
--- a/talk/morelanguage/constexpr.tex
+++ b/talk/morelanguage/constexpr.tex
@@ -1,32 +1,31 @@
 \subsection[cstexpr]{Constant Expressions}
 
 \begin{frame}[fragile]
-  \frametitlecpp[11]{Generalized Constant Expressions}
+  \frametitlecpp[14]{Generalized Constant Expressions}
   \begin{block}{Reason of being}
     \begin{itemize}
-    \item compute constant expressions at compile time
-    \item even if non trivial
+    \item use functions to compute constant expressions at compile time
     \end{itemize}
   \end{block}
   \pause
   \begin{exampleblock}{Example}
     \begin{cppcode*}{}
       constexpr int f(int x) {
-        return x > 1 ? x * f(x - 1) : 1;
-      }
-      constexpr int a = f(5); // computed at compile time
-    \end{cppcode*}
-  \end{exampleblock}
-  \pause
-  \begin{exampleblock}{Example with \cpp14}
-    \begin{cppcode*}{}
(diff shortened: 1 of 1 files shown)


talk/morelanguage/constexpr.tex
```diff
@@ -47,31 +45,67 @@
...
   \begin{block}{Notes}
     \begin{itemize}
-    \item classes can have constexpr member functions
-    \item objects can be constexpr
+    \item A \mintinline{cpp}{constexpr} function \emph{may} be executed at compile time.
+    \begin{itemize}
+      \item Arguments must be \mintinline{cpp}{constexpr} or literals in order to benefit from compile time computation
+      \item Function body must be visible to the compiler
+    \end{itemize}
+    \item A \mintinline{cpp}{constexpr} function can also be used normally
```
hageboeck, line 62:
```suggestion
    \item A \mintinline{cpp}{constexpr} function can also be used at runtime
```

talk/morelanguage/constexpr.tex
```diff
@@ -47,31 +45,67 @@
...
+    \item A \mintinline{cpp}{constexpr} variable must be evaluated at compile time
```
hageboeck, line 63:
```suggestion
    \item A \mintinline{cpp}{constexpr} variable must be initialised at compile time
```

talk/morelanguage/constexpr.tex
```diff
@@ -47,31 +45,67 @@
...
+    \item A constexpr function is implicitly \mintinline{cpp}{inline} (header files)
     \end{itemize}
   \end{block}
 \end{frame}
 
+\begin{frame}[fragile,shrink=15]
+  \frametitlecpp[11]{Limitations of constexpr functions}
+   \begin{alertblock}{C++11}
+     \begin{itemize}
+     \item Non-virtual function with basically a single return statement
```
hageboeck, line 81:
```suggestion
     \item Non-virtual function with a single return statement
```

hageboeck (approved):
Good improvements. I added some more ideas for you to consider.

talk/morelanguage/constexpr.tex
bernhardmgruber: I put the `basically` because you could also have e.g. an alias declaration before the return statement. But yeah, let's just make it simple on the slides.

Issue #173, opened 2022-06-03
Add a slide on `consteval`


bernhardmgruber:
We should mention `consteval` somewhere around `constexpr`, just so people have heard about it.

PR #192, opened 2022-08-02
Add a slide on consteval functions

bernhardmgruber:
Fixes: #173

- [x] Merge #190 first

### DIFF
--- a/talk/morelanguage/constexpr.tex
+++ b/talk/morelanguage/constexpr.tex
@@ -14,18 +14,12 @@
         if (x > 1) return x * f(x - 1);
         return 1;
       }
-      constexpr int a = f(5); // computed at compile time
-      int b = f(n); // computed at runtime
+      constexpr int a = f(5); // computed at compile-time
+      int b = f(5); // maybe computed at compile-time
+      int n = ...;  // runtime value
+      int c = f(n); // computed at runtime
     \end{cppcode*}
   \end{exampleblock}
-  \pause
-  \begin{block}{\cpp11 allowed only a single return statement}
-    \begin{cppcode*}{}
-      constexpr int f(int x) {
-        return x > 1 ? x * f(x - 1) : 1;
-      }
-    \end{cppcode*}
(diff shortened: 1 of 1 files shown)


talk/morelanguage/constexpr.tex
```diff
@@ -14,8 +14,10 @@
         if (x > 1) return x * f(x - 1);
         return 1;
       }
-      constexpr int a = f(5); // computed at compile time
-      int b = f(n); // computed at runtime
+      constexpr int a = f(5); // computed at compile-time
+      int b = f(5); // maybe computed at compile-time
+      int n = ...;  // runtime value
+      int c = f(n); // computed at runtime
```
hageboeck, line 20: This might not fit on the slide, but I like having it here, so we should look how to shrink other things a bit.

talk/morelanguage/constexpr.tex
```diff
@@ -14,18 +14,12 @@
...
     \end{cppcode*}
   \end{exampleblock}
-  \pause
-  \begin{block}{\cpp11 allowed only a single return statement}
-    \begin{cppcode*}{}
-      constexpr int f(int x) {
-        return x > 1 ? x * f(x - 1) : 1;
-      }
-    \end{cppcode*}
-  \end{block}
```
bernhardmgruber, line 28: This is also discussed on the slide about `constexpr` restrictions.

talk/morelanguage/constexpr.tex
```diff
@@ -148,3 +142,27 @@
     static_assert(km_y == 1093, "expected km == 1093 yards!");
   \end{cppcode*}
 \end{frame}
+
+\begin{frame}[fragile]
```
hageboeck, line 146: All good on this slide. Since the four cases are also here, it might be OK to only have two examples in the previous diff.

hageboeck (approved):
With the most recent change, everything should fit on the slide, so good to merge.

Issue #161, opened 2022-03-17
Missing override?


bernhardmgruber:
I don't get this slide:
![image](https://user-images.githubusercontent.com/1224051/158831957-dc371f01-9656-4a33-890a-4121c7650784.png)

The `foo` in `DerivedClass` hides both `foo`s in the `BaseClass`. But the text above says `overriding`. Was it maybe intended to make the `foo`s in `BaseClass` `virtual`?

PR #196, opened 2022-08-03
Add missing virtual

bernhardmgruber:
This slide was meant to show a scenario where a member function is
overriden. Without virtual this was not the case.

Fixes: #161


hageboeck (approved):
LGTM!

Issue #68, opened 2021-11-26
add a slide on CTAD (class template argument deduction)


sponce:
Next to the structured binding slides, we could consider adding a slide on CTAD (class template argument deduction). that works really great with structured bindings and multiple return types in case you don't want to define a struct for the return type:
auto f() { int a; double b; std::string c; ...; return std::tuple{a, b, c}; }
auto [a, b, c] = func();

PR #198, opened 2022-08-03
Add slides on CTAD and deduction guides

bernhardmgruber:
Fixes: #68

### DIFF
--- a/talk/morelanguage/templates.tex
+++ b/talk/morelanguage/templates.tex
@@ -168,6 +168,78 @@
 \end{frame}
 
 \begin{frame}[fragile]
+  \frametitlecpp[17]{Class Template Argument Deduction (CTAD)}
+  \begin{block}{CTAD}
+    \begin{itemize}
+    \item Deduce the template arguments for a class template
+    \item Based on construction arguments
+    \item Only when no template arguments provided
+    \item Since \cpp20: CTAD for aggregates (no constructor needed)
+    \end{itemize}
+  \end{block}
+  \begin{cppcode*}{}
+    template<typename A, typename B, typename C = double>
+    struct Triple {
+      Triple(A a, B b, C c) : a(a), b(b), c(c) {} // C++17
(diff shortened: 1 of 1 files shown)


talk/morelanguage/templates.tex
```diff
@@ -167,6 +167,89 @@
...
+    \end{cppcode*}
+  \end{overprint}
+\end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[17]{Class Template Argument Deduction (CTAD)}
+  \begin{block}{Standard library examples}
+    \begin{cppcode*}{}
+      std::pair p{1.2, true}; // std::pair<double, bool>
+      std::tuple t{1.2, true, 32}; // std::tuple<double, bool, int>
```
hageboeck, line 230: This leaks out to the right. It might not be obvious to everybody that the third is int.

talk/morelanguage/templates.tex
```diff
@@ -167,6 +167,89 @@
...
+      std::vector v{1, 2, 3}; // std::vector<int>
+      std::list l{v.begin(), v.end()}; // std::list<int>
+      std::array a{1, 2, 3}; // std::array<int, 3>
+
+      std::mutex m;
+      std::lock_guard l(m); // std::lock_guard<std::mutex>
+    \end{cppcode*}
+  \end{block}
```
hageboeck, line 250:
```suggestion
```
This seems to be a duplicate.

hageboeck (approved):
This is good after two small fixes.

talk/morelanguage/templates.tex
bernhardmgruber: haha, thx! :D

talk/morelanguage/templates.tex
bernhardmgruber: Oh yeah, I planned to fix that before the commit and then forgot. Thx for spotting!

PR #200, opened 2022-08-08
Refactor Complex in STL exercise

bernhardmgruber:
Since I want to use the `Complex` class from the STL exercise in my follow up modules exercises, I refactored the a bit. Primarily, I wanted to make the code shorter and the operators use hidden friends.

### DIFF
--- a/code/stl/Complex.hpp
+++ b/code/stl/Complex.hpp
@@ -1,74 +1,110 @@
 #include <ostream>
+#include <cmath>
 
 template <typename T=float>
 class Complex_t {
 public:
-    Complex_t(T r, T i);
-    Complex_t();
+    Complex_t() = default;
+    Complex_t(T r, T i) : m_r(r), m_i(i) {}
 
-    T real() const {return m_r;};
-    T imaginary() const {return m_i;};
+    T real() const { return m_r; }
+    T imaginary() const { return m_i; }
 
-    Complex_t operator+(const Complex_t& other) const;
-    Complex_t operator-(const Complex_t& other) const;
-    Complex_t operator*(const Complex_t& other) const;
-    Complex_t operator*(const T factor) const;
-    Complex_t operator/(const T dividend) const;
-    Complex_t& operator+=(const Complex_t& other);
-    bool operator<(const Complex_t& a) const;
+    T norm_sqr() const {
+        return m_r * m_r + m_i * m_i;
+    }
+
+    T norm() const {
+        return std::sqrt(norm_sqr());
+    }
+
+    Complex_t& operator+=(const Complex_t& other) {
+        m_r += other.m_r;
+        m_i += other.m_i;
+        return *this;
+    }
+
+    Complex_t& operator-=(const Complex_t& other) {
+        m_r -= other.m_r;
+        m_i -= other.m_i;
+        return *this;
+    }
+
+    Complex_t& operator*=(const Complex_t& other) {
+        const auto r = m_r * other.m_r - m_i * other.m_i;
+        const auto i = m_r * other.m_i + m_i * other.m_r;
+        m_r = r;
+        m_i = i;
+        return *this;
+    }
+
+    Complex_t& operator*=(T factor) {
+        m_r *= factor;
+        m_i *= factor;
+        return *this;
+    }
+
+    Complex_t& operator/=(const Complex_t& other) {
+        const T ns = other.norm_sqr();
+        const auto r = (m_r * other.m_r + m_i * other.m_i) / ns;
+        const auto i = (m_i * other.m_r - m_r * other.m_i) / ns;
+        m_r = r;
+        m_i = i;
+        return *this;
+    }
+
+    Complex_t& operator/=(T divisor) {
+        m_r /= divisor;
+        m_i /= divisor;
+        return *this;
+    }
+
+    friend Complex_t operator+(Complex_t a, const Complex_t& b) {
+        return a += b;
+    }
+
+    friend Complex_t operator-(Complex_t a, const Complex_t& b) {
+        return a -= b;
+    }
+
+    friend Complex_t operator*(Complex_t a, const Complex_t& b) {
+        return a *= b;
+    }
+
+    friend Complex_t operator*(Complex_t c, T factor) {
+        return c *= factor;
+    }
+
+    friend Complex_t operator*(T factor, Complex_t c) {
+        return c *= factor;
+    }
+
+    friend Complex_t operator/(Complex_t a, Complex_t b) {
+        return a /= b;
+    }
+
+    friend Complex_t operator/(Complex_t c, T divisor) {
+        return c /= divisor;
+    }
+
+    friend Complex_t operator/(T dividend, const Complex_t& c) {
+        return Complex_t(dividend, 0) / c;
+    }
+
+    friend bool operator==(const Complex_t& a, const Complex_t& b) {
+        return a.m_r == b.m_r && a.m_i == b.m_i;
+    }
+
+    friend bool operator<(const Complex_t& a, const Complex_t& b) {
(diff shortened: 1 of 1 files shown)


code/stl/Complex.hpp
```diff
@@ -3,72 +3,51 @@
...
-Complex_t<T>::Complex_t() {}
+    friend Complex_t operator+(Complex_t a, const Complex_t& b) {
+        return a += b;
+    }
 
-template <typename T>
-Complex_t<T>::Complex_t(T r, T i) : m_r(r), m_i(i) {}
+    friend Complex_t operator-(const Complex_t& a, const Complex_t& b) {
+        return Complex_t(a.m_r - b.m_r, a.m_i - b.m_i);
+    }
```
hageboeck, line 24: Not your code initially, but why do we provide `+=` and not `-=`, etc?
Maybe this is unnecessary, indeed, but if you want to reuse it somewhere else, one could imagine to implement `{+,-,*,/}=` and express all of {+,-,*,/} in terms of those.

hageboeck: Are hidden friends in the slides?
Hint: Yes, to explain `swap`, but we don't discuss them. 🙂 
Issue coming ...

code/stl/Complex.hpp
```diff
@@ -3,72 +3,51 @@
...
+                         a.m_r*b.m_i + a.m_i*b.m_r);
+    }
 
-template <typename T>
-Complex_t<T> Complex_t<T>::operator-(const Complex_t<T>& other) const {
-    return Complex_t(m_r - other.m_r, m_i - other.m_i);
-}
+    friend Complex_t operator*(const Complex_t& c, T factor) {
+        return Complex_t(c.m_r*factor, c.m_i*factor);
+    }
```
hageboeck, line 33: Here you removed `Complex*Complex`, or did I miss it? Is that intentional?

hageboeck: Same for `/`

hageboeck (commented):
👍 for the hidden friends. Only two open questions:
- Are we discussing hidden friends? --> No, opening issue.
- Did you want to remove operatorX(Complex, Complex)?

code/stl/Complex.hpp
hageboeck: We don't provide Complex/Complex, but that was the case also before ...

hageboeck: New beef with the `operator*`:
It doesn't compile for `1.f * Complex{}`.

bernhardmgruber: I reproduced the feature set that was there before. But if it makes you feel better, I can add the missing operations :)

code/stl/Complex.hpp
bernhardmgruber: I added the missing operators.

And thx for opening the issue to discuss hidden friends.

code/stl/Complex.hpp
bernhardmgruber: I added the missing operators.

code/stl/Complex.hpp
```diff
@@ -3,72 +3,101 @@
...
-    Complex_t& operator+=(const Complex_t& other);
-    bool operator<(const Complex_t& a) const;
+    Complex_t() = default;
+    Complex_t(T r, T i) : m_r(r), m_i(i) {}
+
+    T real() const { return m_r; }
+    T imaginary() const { return m_i; }
+
+    T norm() const {
+        return m_r * m_r + m_i * m_i;
```
hageboeck, line 13: Here you're actually missing `sqrt`, aren't you?
This function would commonly be called norm2 or normSq.

code/stl/Complex.hpp
```diff
@@ -3,72 +3,101 @@
...
+        return a -= b;
+    }
+
+    friend Complex_t operator*(Complex_t a, const Complex_t& b) {
+        return a *= b;
+    }
+
+    friend Complex_t operator*(Complex_t c, T factor) {
+        return c *= factor;
+    }
```
hageboeck, line 69: Template to kill two birds with one stone, or too complicated?

hageboeck: Three, actually. The below could also be covered.

code/stl/Complex.hpp
```diff
@@ -3,72 +3,101 @@
...
+    friend bool operator==(const Complex_t& a, const Complex_t& b) {
+        return a.m_r == b.m_r && a.m_i == b.m_i;
+    }
+
+    friend bool operator<(const Complex_t& a, const Complex_t& b) {
+        return a.norm() < b.norm();
+    }
+
+    friend std::ostream& operator<<(std::ostream& os, const Complex_t<T>& c) {
+        return os << "(" << c.real() << ", " << c.imaginary() << ")";
```
hageboeck, line 96:
```suggestion
        return os << c.real() << " + " << c.imaginary() << "i";
```

code/stl/Complex.hpp
```diff
@@ -3,72 +3,101 @@
...
+        m_r *= factor;
+        m_i *= factor;
+        return *this;
+    }
+
+    Complex_t& operator/=(const Complex_t& other) {
+        const T n = other.norm();
+        m_i = (m_i * other.m_r - m_r * other.m_i) / n;
+        m_r = (m_r * other.m_r + m_i * other.m_i) / n;
+        return *this;
```
hageboeck, line 46: Sorry, need to leave. Please double check the calculation. Norm above seems to be wrong.

code/stl/Complex.hpp
bernhardmgruber: That's what I thought, but `std::norm` also returns the squared magnitude: https://en.cppreference.com/w/cpp/numeric/complex/norm

code/stl/Complex.hpp
bernhardmgruber: The template would for sure work, but yes, it would complicate the situation a bit, since errors would then be reported one template instantiation deeper. I slightly prefer to leave the code as is.

code/stl/Complex.hpp
bernhardmgruber: There was indeed a bug, but it should be fine now, given that `norm()` returns the squared magnitude.

code/stl/Complex.hpp
```diff
@@ -3,72 +3,103 @@
...
+    Complex_t& operator*=(T factor) {
+        m_r *= factor;
+        m_i *= factor;
+        return *this;
+    }
+
+    Complex_t& operator/=(const Complex_t& other) {
+        const T n = other.norm();
+        const auto r = m_r = (m_r * other.m_r + m_i * other.m_i) / n;
+        const auto i = m_i = (m_i * other.m_r - m_r * other.m_i) / n;
```
hageboeck, line 45:
```suggestion
        const auto r = (m_r * other.m_r + m_i * other.m_i) / n;
        const auto i = (m_i * other.m_r - m_r * other.m_i) / n;
```

hageboeck: Otherwise, you'll use the updated `m_r` to compute `i`. The rest looks correct.

hageboeck (changes requested):

code/stl/Complex.hpp
hageboeck: I find this confusing, but maybe a small comment will clarify it?

code/stl/Complex.hpp
bernhardmgruber: Yeah, that is stupid :D Thx for spotting!

code/stl/Complex.hpp
bernhardmgruber: I renamed it to `norm_sqr()` now and added a new member function `norm()`, which does the sqrt.

hageboeck (approved):

PR #201, opened 2022-08-08
Add exercises for header units and modules

bernhardmgruber:
This PR adds exercises for header units and modules. Since there is no cmake support for both features yet, I could only add Makefiles for g++.

- [x] Merge #200 first
- [x] Merge #222 first

### DIFF
--- a/code/header_units/README.md
+++ b/code/header_units/README.md
@@ -1,0 +1,9 @@
+
+## Instructions
+
+* convert the `#include "Complex.hpp"` to a header unit import
+  * you will need to edit the Makefile and add a new target to precompile `Complex.hpp`
+* convert the standard library includes in `randomize.cpp` to header unit imports
+  * you will need to precompile the standard library headers in your Makefile
+  * with g++-11 and g++-12 you cannot yet precompile `<iostream>` and `<ostream>`, so keep those as a regular includes
+* below g++-12 you will get a linker error when building incrementally, so run `make clean` before each build
--- a/code/header_units/solution/Makefile
+++ b/code/header_units/solution/Makefile
@@ -1,0 +1,16 @@
+all: header_units
+
+clean:
+	rm -rf *.o header_units gcm.cache
+
+algorithm numeric random vector:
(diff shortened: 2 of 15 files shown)


bernhardmgruber:
I especially need someone to read through the Makefiles. This is not my domain :)

code/header_units/README.md
```diff
@@ -0,0 +1,9 @@
+
+## Instructions
+
+* convert the `#include "Complex.hpp"` to a header unit import
+  * you will need to edit the Makefile and add a new target to precompile `Complex.hpp`
+* convert the standard library includes in `randomize.cpp` to header unit imports
+  * you will need to precompile the standard library headers in your Makefile
+  * with g++-11 and g++12 you cannot yet precompile `<iostream>` and `<ostream>`, so keep those as a regular includes
```
hageboeck, line 8: `g++-11` or `g++11`?

code/header_units/solution/Makefile
```diff
@@ -0,0 +1,25 @@
...
+	${CXX} -std=c++20 -fmodules-ts -x c++-system-header $@
+
+numeric:
+	${CXX} -std=c++20 -fmodules-ts -x c++-system-header $@
+
+random:
+	${CXX} -std=c++20 -fmodules-ts -x c++-system-header $@
+
+vector:
+	${CXX} -std=c++20 -fmodules-ts -x c++-system-header $@
```
hageboeck, line 16: Do you want to unify this to one target?

bernhardmgruber: We could, but these can be built in parallel. Can I express this in one target?

code/header_units/README.md
bernhardmgruber: Thx!

code/header_units/solution/Makefile
sponce: I do not see why saying 
```Makefile
algorithm numeric random vector:
	${CXX} -std=c++20 -fmodules-ts -x c++-system-header $@
```
would not be strictly identical to the current code, and in particular would not allow to run in parallel. To be checked though, but it looks like pure syntax.

PR #204, opened 2022-08-09
Hidden friends

### LINKED ISSUE #203: Explain hidden friends.
> Are hidden friends in the slides?

Yes, to explain `swap`, but we don't discuss them. 🙂 

_Originally posted by @hageboeck in https://github.com/hsf-training/cpluspluscourse/pull/200#discussion_r941060535_

hageboeck:
Fix #203 

See also #200 

We are using friends and also hidden friends (+ADL), but it's not explained. I thought we could benefit from one slide on friends related to operators and one for hidden friends as part of ADL.

### PS
The first shows up in the essentials, and both show up in the full course.

### DIFF
--- a/talk/objectorientation/adl.tex
+++ b/talk/objectorientation/adl.tex
@@ -144,3 +144,27 @@
     \end{columns}
   \end{block}
 \end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[98]{ADL and Hidden Friends}
+  \begin{block}{Hidden friend declarations participate in ADL}
+    \begin{itemize}
+      \item Friend functions defined in class scope are ``hidden friends''
+      \item They are \emph{not} member functions (no access to \mintinline{cpp}{this})
+      \item They participate in ADL, e.g.\ \mintinline{cpp}{std::cout << complex}
+    \end{itemize}
+  \end{block}
+  \begin{exampleblock}{\texttt{operator<<} as hidden friend}
+    \begin{cppcode*}{gobble=2}
+      class Complex {
+        float m_real, m_imag; public: Complex(float, float);
+        friend
+        std::ostream & operator<<(std::ostream & os,
+                                  Complex const & c) {
+          return os << c.m_real << " + " << c.m_imag << "i";
+        }
+      };
+      std::cout << Complex{2.f, 3.f}; // Prints 2 + 3i
+    \end{cppcode*}
+  \end{exampleblock}
+\end{frame}
--- a/talk/objectorientation/operators.tex
+++ b/talk/objectorientation/operators.tex
@@ -87,6 +87,32 @@
 \end{frame}
 
 \begin{frame}[fragile]
+  \frametitlecpp[98]{Friend declarations}
+  \begin{block}{Concept}
+    \begin{itemize}
+      \item Functions/classes can be declared \mintinline{cpp}{friend} within a class scope
+      \item They gain access to all private/protected members
+      \item Useful for operators such as \mintinline{cpp}{a + b}
+      \item Don't abuse friends to go around a wrongly designed interface
+      \item Avoid unexpected modifications of class state in a friend
+    \end{itemize}
+  \end{block}
+  \begin{block}{\texttt{operator+} as a friend}
+    \small
+    \begin{cppcode*}{gobble=2}
+      class Complex {
+        float m_real, m_imag;
+        friend
+        Complex operator+(Complex const & a, Complex const & b);
+      };
+      Complex operator+(Complex const & a, Complex const & b) {
+        return Complex{ a.m_real+b.m_real, a.m_imag+b.m_imag };
+      }
+    \end{cppcode*}
+  \end{block}
+\end{frame}
+
+\begin{frame}[fragile]
   \frametitlecpp[98]{Operators}
   \begin{alertblock}{Exercise}
     Write a simple class representing a fraction and pass all tests


talk/objectorientation/adl.tex
```diff
@@ -144,3 +144,27 @@
     \end{columns}
   \end{block}
 \end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[98]{ADL and Hidden Friends}
+  \begin{block}{Hidden friend declarations participate in ADL}
+    \begin{itemize}
+      \item Friends defined in class scope are ``hidden friends''
```
bernhardmgruber, line 152:
```suggestion
      \item Friend functions defined in class scope are ``hidden friends''
```

talk/objectorientation/adl.tex
```diff
@@ -144,3 +144,27 @@
...
+      \item They are \emph{not} class members (no access to \mintinline{cpp}{this})
```
bernhardmgruber, line 153:
```suggestion
      \item They are \emph{not} member functions (no access to \mintinline{cpp}{this})
```

talk/objectorientation/adl.tex
```diff
@@ -144,3 +144,27 @@
...
+      \item They participate in ADL, e.g.\ \mintinline{cpp}{std::cout << complex}
+    \end{itemize}
+  \end{block}
+  \begin{exampleblock}{Ostream operator as hidden friend}
```
bernhardmgruber, line 157: Can you please try to make the title more precise? You could also mean `operator>>` here. Maybe:
```suggestion
  \begin{exampleblock}{operator<< as hidden friend}
```
`operator<<` is sometimes also called the "put operator".

talk/objectorientation/adl.tex
```diff
@@ -144,3 +144,27 @@
...
+    \begin{cppcode*}{gobble=2,linenos=false}
```
bernhardmgruber, line 158: What's wrong with line numbers here? I like them, so I can refer to them when talking.

talk/objectorientation/adl.tex
```diff
@@ -144,3 +144,27 @@
...
+      class Complex {
+        float m_real, m_imag; // + Complex(float, float) ...
+        friend
+        std::ostream & operator<<(std::ostream & os,
+                                  Complex const & c) {
+          return os << c.m_real << " + " << c.m_imag << "i";
+        }
+      };
+      std::cout << Complex(2.f, 3.f); // Prints 2 + 3i
```
bernhardmgruber, line 167: You could make it an aggregate :)
```suggestion
      struct Complex {
        float m_real, m_imag;
        friend
        std::ostream & operator<<(std::ostream & os,
                                  Complex const & c) {
          return os << c.m_real << " + " << c.m_imag << "i";
        }
      };
      std::cout << Complex{2.f, 3.f}; // Prints 2 + 3i
```
Feel free to ignore this one.

talk/objectorientation/operators.tex
```diff
@@ -86,6 +86,30 @@
   \end{block}
 \end{frame}
 
+\begin{frame}[fragile]
+  \frametitlecpp[98]{Friend declarations}
+  \begin{block}{Concept}
+    \begin{itemize}
+      \item Functions/classes can be declared \mintinline{cpp}{friend} within a class scope
+      \item They gain access to all private members
```
bernhardmgruber, line 94:
```suggestion
      \item They gain access to all private/protected members
```

talk/objectorientation/operators.tex
```diff
@@ -86,6 +86,30 @@
...
+      \item Useful for operators such as \mintinline{cpp}{a + b}
+      \item Don't abuse this! Don't modify an object in a friend!
```
bernhardmgruber, line 96: > Don't modify an object in a friend!

What is the rational for this? Why shouldn't I grant a different class the ability to modify me?

talk/objectorientation/operators.tex
```diff
@@ -86,6 +86,30 @@
...
+    \end{itemize}
+  \end{block}
+  \begin{block}{operator+ as a friend}
+    \begin{cppcode*}{gobble=2,linenos=false}
```
bernhardmgruber, line 100: I would prefer line numbers :)

talk/objectorientation/operators.tex
```diff
@@ -86,6 +86,30 @@
...
+      struct Complex {
+        float m_real; float m_imag;
+        friend Complex operator+(Complex & a, Complex & b);
+      };
+
+      Complex operator+(Complex & a, Complex & b) {
+        return Complex(a.m_real+b.m_real,a.m_imag+b.m_imag);
+      }
```
bernhardmgruber, line 108:
```suggestion
      struct Complex {
        float m_real; float m_imag;
        friend Complex operator+(Complex & a, Complex & b);
      };

      Complex operator+(Complex & a, Complex & b) {
        return Complex(a.m_real+b.m_real,a.m_imag+b.m_imag);
      }
```
I think you miss `const` on the parameters.
And the point of friendship would be stronger, if `m_real` and `m_imag` were private.

talk/objectorientation/adl.tex
hageboeck: The stuff didn't fit on the slide. I'll try again.

talk/objectorientation/adl.tex
hageboeck: I know. But if I do, I don't need to be a friend. 😉

One thing about the `{ }`. I believe that we're not at `{ }` yet. However, these days, we actually want the `{ }`, so should we use them as much as makes sense even before we explain the details?

talk/objectorientation/operators.tex
hageboeck: Of course you can. And there's cases where it might make sense. But more often than not, I see the friends being abused as a convenient way out of a proper refactoring. Especially in HEP in older codes.
What I want to say is:
- They are great for operators where you mostly need read access.
- If you use write access, all bets are off. Don't do stupid things / unexpected things / thread-unsafe things / ...

I didn't know how to say this in one sentence.

talk/objectorientation/adl.tex
bernhardmgruber: > I know. But if I do, I don't need to be a friend.

Right. Ignore my comment then :)

> One thing about the { }. I believe that we're not at { } yet. However, these days, we actually want the { }, so should we use them as much as makes sense even before we explain the details?

Yes, I would use them.

talk/objectorientation/operators.tex
hageboeck: The const doesn't fit on the slide, so I removed it. Also, at this point, we haven't explained constness yet. OK to leave it out or shrink the text?

hageboeck: I'll check it private fits ...

talk/objectorientation/operators.tex
bernhardmgruber: Hmm, I see. Then maybe you can add:
```suggestion
      \item Don't abuse this! Avoid modifying an object in a friend, use the objects member functions instead.
```
Maybe also split the two sentences into two bullet points.

talk/objectorientation/operators.tex
bernhardmgruber: C++ is so interdependent, I think we cannot topsort all the content to only use features that have been previously explained. And I would like to avoid having code on the slides that people should not write in practice, like an operator with a mutable argument. So yeah, I would prefer shrinking the slide to not having `const` :)

talk/objectorientation/operators.tex
hageboeck: They now read:
```diff
+      \item Don't abuse friends to go around a wrongly designed interface
+      \item Avoid unexpected modifications of class state in a friend
```

talk/objectorientation/operators.tex
hageboeck: Manage to fit everything.

talk/objectorientation/operators.tex
```diff
@@ -86,6 +86,32 @@
...
+      \item Useful for operators such as \mintinline{cpp}{a + b}
+      \item Don't abuse friends to go around a wrongly designed interface
+      \item Avoid unexpected modifications of class state in a friend
+    \end{itemize}
+  \end{block}
+  \begin{block}{operator+ as a friend}
+    \small
+    \begin{cppcode*}{gobble=2}
+      class Complex {
+        float m_real, m_imag; public: Complex(float, float);
```
bernhardmgruber, line 104: Since you are not constructing a `Complex` on this slide, you could omit the constructor if you want:
```suggestion
        float m_real, m_imag;
```

talk/objectorientation/operators.tex
```diff
@@ -86,6 +86,32 @@
...
+  \begin{block}{Concept}
+    \begin{itemize}
+      \item Functions/classes can be declared \mintinline{cpp}{friend} within a class scope
+      \item They gain access to all private/protected members
+      \item Useful for operators such as \mintinline{cpp}{a + b}
+      \item Don't abuse friends to go around a wrongly designed interface
+      \item Avoid unexpected modifications of class state in a friend
+    \end{itemize}
+  \end{block}
+  \begin{block}{operator+ as a friend}
```
bernhardmgruber, line 100: On the other slide, you used a different font:
```suggestion
  \begin{block}{\texttt{operator+} as a friend}
```

bernhardmgruber (approved):
Looks good! I have two more suggestions if you want.

talk/objectorientation/operators.tex
hageboeck: That's because your suggestion to use `operator<<` was typeset as `operator¡¡`. I unified it now.

bernhardmgruber (approved):

talk/objectorientation/operators.tex
```diff
@@ -86,6 +86,32 @@
...
+      \item Useful for operators such as \mintinline{cpp}{a + b}
+      \item Don't abuse friends to go around a wrongly designed interface
+      \item Avoid unexpected modifications of class state in a friend
+    \end{itemize}
+  \end{block}
+  \begin{block}{\texttt{operator+} as a friend}
+    \small
+    \begin{cppcode*}{gobble=2}
+      class Complex {
+        float m_real, m_imag;
```
sponce, line 104: This does not compile if you do not declare a constructor
```cpp
Complex( float, float );
```

sponce: Fixed in #229

bernhardmgruber: @hageboeck originally had ` // + Complex(float, float) ...` on that line as a comment, and I suggested to make `Complex` a `struct`. We must have missed this in the end.

Issue #71, opened 2021-11-26
Rework introductory examples of RAII


sponce:
from @bernhardmgruber :
the examples are too much C, people should not be reminded again that malloc, free, strncpy, FILE, fopen, fputs, fclose, EOF exist. maybe we can find a C++ version that news an array for processing some numbers, the FILE example is actually good, because it shows that C++ can do better than C here. But I am afraid that students will remember these codes as something they can use. Like we show FILE* a couple of times but never std::ifstream, so what will students remember when they need to read a file?

chavid:
+1

PR #208, opened 2022-08-10
Refactor examples to use less C

bernhardmgruber:
Addresses part of #71.

### DIFF
--- a/code/condition_variable/condition_variable.cpp
+++ b/code/condition_variable/condition_variable.cpp
@@ -13,7 +13,7 @@
  */
 using namespace std::chrono_literals; // We can write 1s
 
-std::mutex cout_mutex{}; // We need this to synchronise printing
+std::mutex cout_mutex; // We need this to synchronise printing
 
 // Print contents of the stream to cout in a thread-safe manner.
 // This class consumes stream inputs, buffers them, and writes them
@@ -61,7 +61,7 @@ bool process(unsigned int threadIdx, Data const & data) {
 
   // Burn some CPU cycles to simulate intensive data processing.
   const auto startTime = std::chrono::high_resolution_clock::now();
-  int dummyCounter = 0;
+  unsigned dummyCounter = 0;
   while (std::chrono::high_resolution_clock::now() - startTime < 5s) {
     ++dummyCounter;
   }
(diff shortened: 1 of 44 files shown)


hageboeck (approved):
This looks all good, but it would have been easier to have a few commits. E.g. the move to  C++17 could have been one commit, so I don't have 44 diffs to compare in one go. 🙂

hageboeck:
PS:
I don't understand why you move the loop counter from `unsigned int` to `int` in one place, and in another you do the inverse. 😄  It doesn't make a difference, but I'm interested in the thought process.

code/condition_variable/condition_variable.cpp
```diff
@@ -61,7 +61,7 @@ bool process(unsigned int threadIdx, Data const & data) {
 
   // Burn some CPU cycles to simulate intensive data processing.
   const auto startTime = std::chrono::high_resolution_clock::now();
-  int dummyCounter = 0;
+  unsigned dummyCounter = 0;
   while (std::chrono::high_resolution_clock::now() - startTime < 5s) {
     ++dummyCounter;
   }
```
bernhardmgruber, line 67: Needs to be unsigned, because it can intentionally overflow and should wrap around. Signed int overflow is UB.

code/smartPointers/smartPointers.cpp
```diff
@@ -252,7 +252,7 @@ class Owner {
 void problem4_1() {
     std::vector<Owner> owners;
 
-    for (unsigned int i=0; i < 5; ++i) {
+    for (int i=0; i < 5; ++i) {
```
bernhardmgruber, line 255: Changed to `int` because you compare `i < 5`, where the second operand is also an `int`. But honestly, it does not matter :) `i` is also not used in the body of the loop.

bernhardmgruber:
> This looks all good, but it would have been easier to have a few commits.

That's what I had initially, one commit per exercise. But then I had to fix some more things and rebasing the changes into the history became annoying. So I decided to squash.

> E.g. the move to C++17 could have been one commit, so I don't have 44 diffs to compare in one go. slightly_smiling_face

All examples are on C++17 by default with cmake. Only some of the Makefiles don't specify C++17.

PR #216, opened 2022-08-11
Use less C in the RAII introduction slides

bernhardmgruber:
Fixes: #71

### DIFF
--- a/talk/morelanguage/raii.tex
+++ b/talk/morelanguage/raii.tex
@@ -7,11 +7,11 @@
     \begin{cppcode*}{xleftmargin=20pt}
       char *s;
       try {
-        callThatThrows();
-        s = (char*) malloc(...);
-        strncpy(s, ...);
+        foo(); // may throw
+        s = new char[100];
+        read_line(s);
       } catch (...) { ... }
-      bar(s);
+      process_line(s);
     \end{cppcode*}
   \end{exampleblock}
   \pause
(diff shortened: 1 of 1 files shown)


talk/morelanguage/raii.tex
```diff
@@ -47,16 +47,19 @@
...
-      try {
-        FILE *handle = std::fopen(path, "w+");
-        if (nullptr == handle) { throw ... }
-        if (std::fputs(str, handle) == EOF) {
-          throw ...
-        }
-        fclose(handle);
-      } catch (...) { ... }
+      std::FILE *handle = std::fopen(path, "w+");
+      if (nullptr == handle) { throw ... }
```
hageboeck, line 51:
```suggestion
      if (handle == nullptr) { throw ... }
```
I don't insist, but isn't this easier to read?

hageboeck (approved):
LGTM, with two additions:
- There's a mostly formatting-related PR to your branch that you could please merge.
- I have one code suggestion that not very important, though.

talk/morelanguage/raii.tex
bernhardmgruber: There is this clan of C++ developers that believes putting an rvalue on the LHS of an equality comparison detects accidental misspelling the operator as assignment:
```c++
if (nullptr = handle) // compiler error
if (handle = nullptr) // compiles fine (but may warn)
```
I think recent compilers warn on this anyway and you should put parentheses around it to silence the warnings:
```c++
if (handle = nullptr) // may warn
if ((handle = nullptr)) // does not warn
```

sponce: Yes. This is the origin of this. An LHCb old convention, put in place 15-20 years before compilers added the warnings :-) Now I'm not sure it's really useful nowadays and this example should not be polluted by such considerations, so I woudl keep the change

PR #217, opened 2022-08-11
Minor slide improvements

bernhardmgruber:
I read again through all the slides and here are some small changes I would propose.

### DIFF
--- a/talk/expert/cpp20spaceship.tex
+++ b/talk/expert/cpp20spaceship.tex
@@ -1,4 +1,4 @@
-\subsection[spaceship]{The <=> operator}
+\subsection[spaceship]{The \texttt{<=>} operator}
 
 \begin{frame}[fragile]
   \frametitlecpp[20]{The three-way comparison operator}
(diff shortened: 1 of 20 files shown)


talk/expert/cpp20spaceship.tex
```diff
@@ -114,7 +114,9 @@
   \end{exampleblock}
```
hageboeck, line 114: Line https://github.com/hsf-training/cpluspluscourse/pull/217/files#diff-a6dec9ca6d8f983ed52847bf233bc288a103be99d342571e6717bf548e042837R105 above leaks out of the box, but I'm not allowed to comment there.

hageboeck: Next commit will fix that.

talk/expert/perfectforwarding.tex
```diff
@@ -135,22 +138,23 @@
       void func(T&& t) {}
     \end{cppcode*}
   \end{block}
-  Next to a template parameter, \mintinline{cpp}{&&} is not an rvalue, but a ``forwarding reference'' (aka. ``universal reference'')\\
-  \mintinline{cpp}{T&&} actual type depends on the arguments passed to func
   \begin{itemize}
-  \item if an lvalue of type U is given, T is deduced to \mintinline{cpp}{U&}
-  \item otherwise, collapse references normally
+    \item Next to a template parameter, \mintinline{cpp}{&&} is not an rvalue, but a ``forwarding reference'' (aka. ``universal reference'')
```
hageboeck, line 142:
```suggestion
    \item Next to a template parameter, \mintinline{cpp}{&&} is not an rvalue, but a ``forwarding reference'' (aka.\ ``universal reference'')
```

talk/expert/sfinae.tex
```diff
@@ -191,15 +191,15 @@
       using enable_if_t = typename enable_if<B,T>::type;
     \end{cppcode*}
     \begin{itemize}
-    \item If B is true, has a alias \texttt{type} to type T
-    \item otherwise, has no \texttt{type} alias
+    \item If \mintinline{cpp}{B} is true, has a alias \mintinline{cpp}{type} to type \mintinline{cpp}{T}
```
hageboeck, line 194:
```suggestion
    \item If \mintinline{cpp}{B} is true, has an alias \mintinline{cpp}{type} to type \mintinline{cpp}{T}
```

hageboeck (approved):
I like it a lot. Good work. I have found even more things, and I'll try to push a commit directly to this branch now. 🤞

bernhardmgruber:
What should I do here about your suggestions? Commit them or will you add commits?

hageboeck:
> What should I do here about your suggestions? Commit them or will you add commits?

I pushed a commit to take care of those.

PR #220, opened 2022-08-12
Add slides on assertions.

### LINKED ISSUE #156: Should we discuss assert?
Assert is mentioned on a few slides, but we don't discuss it. That would be something for good practices.

hageboeck:
Fix #156.

## For reviewers:
I was unsure as to where to put this. It now sits at the end of basic concepts. Any better ideas?

### DIFF
--- a/talk/basicconcepts/assert.tex
+++ b/talk/basicconcepts/assert.tex
@@ -1,0 +1,94 @@
+\subsection[assert]{Assertions}
+
+\begin{frame}[fragile]
+  \frametitlecpp[98]{Assertions}
+  \begin{block}{Checking invariants in a program}
+    \begin{itemize}
+      \item An invariant is a property that is guaranteed to be true during certain phases of a program, and the program might crash or yield wrong results if it is violated
+      \begin{itemize}
+        \item ``Here, `a' should always be positive''
+      \end{itemize}
+      \item This can be checked using \mintinline{cpp}{assert}
+      \item The program will be aborted if the assertion fails
+    \end{itemize}
+  \end{block}
+  \begin{exampleblock}{assert in practice}
+    \begin{overprint}
+    \onslide<1>
+    \begin{cppcode*}{}
+      #include <cassert>
+      double f(double a) {
+        // [...] do stuff with a
+        // [...] that should leave it positive
+        assert(a > 0.);
+        return std::sqrt(a);
+      }
+    \end{cppcode*}
+    \onslide<2>
+    \begin{Verbatim}
+% ./testAssert
+Assertion failed: (a > 0.), function f,
+                  file testAssert.cpp, line 5.
+Abort trap: 6
+    \end{Verbatim}
+    \end{overprint}
+  \end{exampleblock}
+\end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[98]{Assertions}
+  \begin{block}{Good practices}
+    \begin{itemize}
+      \item Assertions are mostly for developers and debugging
+      \item Use them to check important invariants of your program
+      \item Prefer handling user-facing errors with helpful error messages/exceptions
+      \item Assertions can impact the speed of a program
+      \begin{itemize}
+        \item Assertions are disabled when the macro \mintinline{cpp}{NDEBUG} is defined
+        \item Decide if you want to disable them when you release code
+      \end{itemize}
+    \end{itemize}
+  \end{block}
+  \begin{exampleblock}{Disabling assertions}
+    \small
+    Compile a program with NDEBUG defined:\\
+    \texttt{g++ \textcolor{blue}{-DNDEBUG} -O2 -W[...] test.cpp -o test.exe}
+
+    \begin{cppcode*}{}
+      double f(double a) {
+        assert(a > 0.); // no effect
+        return std::sqrt(a);
+      }
+    \end{cppcode*}
+  \end{exampleblock}
+\end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[17]{Static Assert}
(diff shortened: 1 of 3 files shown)


talk/basicconcepts/assert.tex
```diff
@@ -0,0 +1,89 @@
...
+  \frametitlecpp[98]{Assertions}
+  \begin{block}{Good practices}
+    \begin{itemize}
+      \item Assertions are for developers and debugging; not for users
+      \item Use them to check important invariants of your program
+      \item Handle user-facing errors with error messages/exceptions
+      \item Assertions can impact the speed of a program. Therefore:
+      \begin{itemize}
+        \item Assertions are disabled when the macro \mintinline{cpp}{NDEBUG} is defined
+        \item Make sure it is defined when you release code
```
bernhardmgruber, line 38: I think this is generally the norm. However, my company decided in the past to also ship release builds with assertions enabled, because we would rather prefer a program termination to corrupted data.

talk/basicconcepts/assert.tex
```diff
@@ -0,0 +1,89 @@
...
+        \item Many build systems do this automatically
+      \end{itemize}
+    \end{itemize}
+  \end{block}
+  \begin{exampleblock}{Disabling assertions}
+    \begin{overprint}
+    \onslide<1>
+    \begin{cppcode*}{}
+      #define NDEBUG
+      #include <cassert>
```
bernhardmgruber, line 48: I dislike when people configure headers like this. Such macros should really be defined at the build system level (which you show as well). Should we point out that the latter option is to be preferred?

talk/basicconcepts/assert.tex
```diff
@@ -0,0 +1,89 @@
...
+    \end{cppcode*}
+    \onslide<2>
+    Compile entire project with NDEBUG:\\
+    \texttt{g++ \textcolor{blue}{-DNDEBUG} -O2 -W[...] test.cpp -o test.exe}
+    \end{overprint}
+  \end{exampleblock}
+\end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[17]{Static Assert}
```
bernhardmgruber, line 62: We already have slides on `static_assert` in `talk/morelanguage/constexpr.tex` :) Please unify!

talk/basicconcepts/assert.tex
```diff
@@ -0,0 +1,89 @@
+\subsection[assert]{Assertions}
+
+\begin{frame}[fragile]
+  \frametitlecpp[98]{Assertions}
+  \begin{block}{Checking invariants in a program}
+    \begin{itemize}
+      \item An invariant is a property of a program that is always true, and the program might crash or yield wrong results if violated
+      \begin{itemize}
+        \item ``Here, `a' should always be positive''
```
bernhardmgruber, line 9: Isn't this a contradiction? An invariant is supposed to be true during the entire duration of a program. But the quote says: `Here, ...`.

I am also unsure whether the term invariant is actually correct. An assertion tests a condition at a specific point in program execution. The condition may change immediately after the assertion or have been different before.

bernhardmgruber:
> I was unsure as to where to put this. It now sits at the end of basic concepts. Any better ideas?

I find it useful in the beginning as well, so we can use assertions on the slides following up.

talk/basicconcepts/assert.tex
hageboeck: This is an example on cppreference. I also don't like it. Lets remove it entirely?

talk/basicconcepts/assert.tex
hageboeck: I wouldn't say it is. I don't have an authoritative academic source at hand, but e.g. on [wikipedia](https://en.wikipedia.org/wiki/Invariant_(mathematics)#Invariants_in_computer_science), emphasis mine:
> In [computer science](https://en.wikipedia.org/wiki/Computer_science), an invariant is a [logical assertion](https://en.wikipedia.org/wiki/Logical_assertion) **that is always held to be true during a certain phase of execution of a [computer program](https://en.wikipedia.org/wiki/Computer_program)**. For example, a [loop invariant](https://en.wikipedia.org/wiki/Loop_invariant) is a condition that is true at the beginning and the end of every execution of a loop.
>
> [...]
>
> Programmers often use [assertions](https://en.wikipedia.org/wiki/Assertion_(computing)) in their code to make invariants explicit. Some [object oriented](https://en.wikipedia.org/wiki/Object_oriented) [programming languages](https://en.wikipedia.org/wiki/Programming_language) have a special syntax for specifying [class invariants](https://en.wikipedia.org/wiki/Class_invariant).

talk/basicconcepts/assert.tex
hageboeck: This will be watered down in the next push.

talk/basicconcepts/assert.tex
bernhardmgruber: Great, let's remove it then or at least "grey it out" or something.

talk/basicconcepts/assert.tex
hageboeck: I wrote a bit about the fact that constexpr powers static_assert, but I wonder if I shouldn't just remove the slide. Opinions?

bernhardmgruber: I think I would remove the `static_assert` slide in the constexpr section, and just add here that the condition must be a constant expression.

hageboeck: OK, done. The fact that it has to be `constexpr` has been mentioned. @bernhardmgruber

bernhardmgruber: Great!

bernhardmgruber (approved):

PR #222, opened 2022-08-12
Improve STL exercise

bernhardmgruber:
* Prefix STL calls with std::
* Add const to variables with are not mutated
* Add lambda based alternative solution

### DIFF
--- a/code/stl/randomize.cpp
+++ b/code/stl/randomize.cpp
@@ -17,7 +17,7 @@ void compute(int len, T initial, T step) {
     // compute differences
     std::adjacent_difference(...);
 
-    // compute standard deviation of it
+    // compute standard deviation of all differences
     const T sum = std::reduce(...);
     const T sumsq = std::reduce(..., [](...) { ...; });
     const T mean = sum/len;
--- a/code/stl/randomize.cpp
+++ b/code/stl/randomize.cpp
@@ -5,25 +5,23 @@
 #include <random>
 #include "Complex.hpp"
 
-using namespace std;
-
 template<typename T>
 void compute(int len, T initial, T step) {
     // allocate vectors
     std::vector<T> v(len+1), diffs(len+1);
 
     // fill and randomize v
-    generate(, , [](...) { ...; });
-    shuffle(..., std::default_random_engine{});
+    std::generate(, , [](...) { ...; });
+    std::shuffle(..., std::default_random_engine{});
 
     // compute differences
-    adjacent_difference(...);
+    std::adjacent_difference(...);
 
     // compute standard deviation of it
-    T sum = reduce(...);
-    T sumsq = reduce(..., [](...) { ...; });
-    T mean = sum/len;
-    T variance = sumsq/len - mean*mean;
+    const T sum = std::reduce(...);
+    const T sumsq = std::reduce(..., [](...) { ...; });
+    const T mean = sum/len;
+    const T variance = sumsq/len - mean*mean;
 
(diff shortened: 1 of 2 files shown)


code/stl/solution/randomize.sol.cpp
```diff
@@ -29,17 +27,27 @@ void compute(int len, T initial, T step) {
...
-    adjacent_difference(v.begin(), v.end(), diffs.begin());
+    std::adjacent_difference(v.begin(), v.end(), diffs.begin());
 
     // compute standard deviation of it
-    T sum = reduce(diffs.begin()+1, diffs.end(), T());
-    T sumsq = reduce(diffs.begin()+1, diffs.end(), T(), sumsquare<T>());
-    T mean = sum/len;
-    T variance = sumsq/len - mean*mean;
+    const T sum = std::reduce(diffs.begin()+1, diffs.end(), T());
+    const T sumsq = std::reduce(diffs.begin()+1, diffs.end(), T(), sumsquare<T>());
```
hageboeck, line 45:
```suggestion
    const T sum = std::reduce(diffs.begin(), diffs.end(), T());
    const T sumsq = std::reduce(diffs.begin(), diffs.end(), T(), sumsquare<T>());
```
Why do you skip the first element? Was this meant to be
```c++
    const T sum = std::reduce(diffs.begin(), diffs.end(), diffs[0]);
```
?

hageboeck (changes requested):
I might have found a bug, please check.

code/stl/solution/randomize.sol.cpp
bernhardmgruber: I think this is correct, since `std::adjacent_difference` only writes differences starting at element 1. So `diffs[0]` is just `v[0]` and `diffs[1]` is the first difference (`v[1] - v[0]`).

code/stl/solution/randomize.sol.cpp
```diff
@@ -29,17 +27,27 @@ void compute(int len, T initial, T step) {
...
+    //     return cur;
+    // });
+
+    std::shuffle(v.begin(), v.end(), std::default_random_engine{});
 
     // compute differences
-    adjacent_difference(v.begin(), v.end(), diffs.begin());
+    std::adjacent_difference(v.begin(), v.end(), diffs.begin());
 
     // compute standard deviation of it
```
hageboeck, line 43:
```suggestion
    // compute standard deviation of all differences.
    // Note that the first element is just the original element itself, so we need to skip it.
```

code/stl/solution/randomize.sol.cpp
hageboeck: Sorry, I ignored where those values came from. This is correct, but not very intuitive without any explanation.

code/stl/randomize.cpp
```diff
@@ -5,25 +5,23 @@
...
-    generate(, , [](...) { ...; });
-    shuffle(..., std::default_random_engine{});
+    std::generate(, , [](...) { ...; });
+    std::shuffle(..., std::default_random_engine{});
 
     // compute differences
-    adjacent_difference(...);
+    std::adjacent_difference(...);
 
     // compute standard deviation of it
```
hageboeck, line 20:
```suggestion
    // compute standard deviation of all differences
```

code/stl/solution/randomize.sol.cpp
bernhardmgruber: With your suggestions (that I committed), this is resolved now I guess?

hageboeck (approved):

PR #223, opened 2022-08-12
Refactor basic/advanced tex file inclusions

bernhardmgruber:
I have contemplated such a refactoring twice now and decided to draft a small prototype now to request some comments.

This PR:
    * Add new comment environment advanced
    * Replace \ifthenelse by latex built-in \if*** and the advanced comment
      environment, because it handles verbatim correctly
    * The command to put into onlybasics.tex now changed to \basictrue

- [x] Depends on #248

### DIFF
--- a/talk/C++Course.tex
+++ b/talk/C++Course.tex
@@ -4,9 +4,18 @@
 
 \input{setup}
 
-\newboolean{onlybasics}
-\setboolean{onlybasics}{false}
+% setup for the basic/advanced course switch
+% create a new latex if called basic (false by default)
+\newif\ifbasic
+%\basictrue % uncomment to make basic true
 \IfFileExists{onlybasics.tex}{\input{onlybasics}}{}
+
+% create a comment environment advanced. depending on the value of basic, we exclude or include it.
+\ifbasic
+\excludecomment{advanced}
+\else
+\includecomment{advanced}
+\fi
 
 \newunicodechar{σ}{$\sigma$}
 \newunicodechar{µ}{$\mu$}
@@ -99,19 +108,7 @@
 \input{basicconcepts/inline}
 \input{basicconcepts/assert}
 
-% basic version has subset of tools at this stage
-\ifthenelse{\boolean{onlybasics}}{
-  \section[Tool]{Useful tools}
-  \input{tools/editors}
-  \input{tools/vcs}
-  \input{tools/formatting}
-  \input{tools/compiling}
-  \input{tools/debugging}
-}{}
-
-% basic version only keeps a subset of these 2 chapters
 \section[OO]{Object orientation (OO)}
-\ifthenelse{\boolean{onlybasics}}{
 \input{objectorientation/objectsclasses}
 \input{objectorientation/inheritance}
 \input{objectorientation/constructors}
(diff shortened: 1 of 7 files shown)


hageboeck:
I'm in favour of some kind of refactoring. I suggest to bring this up in the meeting on Wed.

bernhardmgruber:
I still wonder how to deal with fragments like this:
```latex
\ifthenelse{\boolean{onlybasics}}{%
\begin{frame}<handout:0|beamer:0>[fragile]
}{%
\begin{frame}[fragile]
}
```
(that is on the slide for the `final` keyword on classes)
Any ideas how to mark this nicely as advanced?

hageboeck:
> I still wonder how to deal with fragments like this:
> 
> ```
> \ifthenelse{\boolean{onlybasics}}{%
> \begin{frame}<handout:0|beamer:0>[fragile]
> }{%
> \begin{frame}[fragile]
> }
> ```
> 
> (that is on the slide for the `final` keyword on classes) Any ideas how to mark this nicely as advanced?

What's the question exactly? This if disables the frame by specifying that it's not going to be shown.

hageboeck:
> What's the question exactly? This if disables the frame by specifying that it's not going to be shown.

Ah, I should read better. 🙂
I have a few ideas how to "mark" it, but all not so great:
- Make an index of advanced slides that are not part of any advanced chapter. That's going to be pretty chaotic to lecture.
- Move the slide to an advanced chapter
- Add a label to the title

bernhardmgruber:
> * Make an index of advanced slides that are not part of any advanced chapter. That's going to be pretty chaotic to lecture.

That sounds messy :)

> * Move the slide to an advanced chapter

Hmm. That would disconnect the slides from the chapters they belong to. Like the slide on `final` logically belongs into the OO section. I don't think we should move it into a different section/chapter.

> * Add a label to the title

Possibly. I was recently also thinking on whether we should just change the background color. Or maybe just gray out some slides. Like: When we build the advanced course, all basic slides are just grayed-out. So we know that we can skip over them, but we still have them if we want to explain something on them.

bernhardmgruber:
So, I eventually found a way with `\newif` to make the toggling of basic/advanced slides work, since they contain verbatim environments. We now have a new environment `advanced` that you can wrap around any advanced content and a new command `\ifadvanced{...}` with the same idea. The latter does not support verbatim arguments, but the environment does.

talk/C++Course.tex
```diff
@@ -4,10 +4,22 @@
...
+\else
+\includecomment{advanced}
+\fi
+
+\newcommand\ifadvanced[1]{
+\ifbasic\else
+#1
+\fi
+}
+
```
hageboeck, line 22: I don't know if it was just a misunderstanding, but I struggled a lot to understand what you're trying to to here. Before merging, I suggest to add some comments as to what's going on here.

Now, the biggest problem was to first read up on the comment package (which is only included at the very bottom of the PR -- not your fault), and then matching what I read there to what's going on here.
Now that I think I know what's going on, the biggest question is:
- Why are you declaring two ways of commenting in the advanced part?

Unless there's a latex reason I don't see, isn't it possible to do *everything* using `advanced` and `endadvanced`? E.g.
```latex
\begin{advanced}
  ...
\end{advanced}
```

Why is there a command `\ifadvanced` that's kind of doing the same thing?

Also, I found
```latex
\ifbasic\else
```
very confusing.
Maybe this is easier to read?
```latex
\ifbasic
%
\else
#1
\fi
```
or `\relax` etc. What do you think?

hageboeck: PS:
I want to add that a really big advantage of the new method is that the `\input` duplication below goes away. I'm much in favour of this!

talk/setup.tex
```diff
@@ -80,7 +80,7 @@
 \usepackage{tikz-uml}
 
 \usepackage{booktabs}
-\usepackage{ifthen}
+\usepackage{comment}
```
hageboeck, line 83: I think this will break the PR about the index of good practices and index of exercises etc. That's not a showstopper, though. I can re-introduce `ifthen` if necessary.

bernhardmgruber: I had to switch over to the commend package for including/excluding content, because it can handle verbatim environments inside the commented section.

talk/C++Course.tex
bernhardmgruber: I added a few comments to explain what I am trying to achieve.

> Why are you declaring two ways of commenting in the advanced part?

Because I thought having a command that just wraps the advanced part in parentheses would look nicer in some places. But you are correct, that it is just a duplication of functionality. I removed it now and handle everything using the `advanced` environment.

talk/setup.tex
bernhardmgruber: You can still use the `ifthen` package if you want! Just readd it in your PR!

hageboeck (approved):
LGTM!

PR #225, opened 2022-08-15
Publish slides from CI

bernhardmgruber:
This is a modification of #184.

### DIFF
--- a/.github/workflows/slides.yml
+++ b/.github/workflows/slides.yml
@@ -1,0 +1,47 @@
+name: Publish slides
+on:
+  push:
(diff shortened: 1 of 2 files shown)


bernhardmgruber:
I tried the action a couple of times, but the workflow did not have permissions to push again to the download branch. I might just as well be, that we need to merge this action in, so it originates from the same repository.

klieret:
Could you update the readme accordingly? Regarding how to get the latest version?

bernhardmgruber:
This is #228, waiting to be merged.

Issue #106, opened 2022-03-09
Find better example code for move semantics.


hageboeck:
And copy elision will take care of `MyVector` ...
Maybe we should think about something for the future.

_Originally posted by @hageboeck in https://github.com/hsf-training/cpluspluscourse/pull/103#discussion_r822684920_

The problem is that all the code we suggest will get optimised. We should come up with something that really improves only with move semantics.

bernhardmgruber:
As I said in the comments, NRVO is not guaranteed:

```c++
A f() {
  A a;
  ...
  return a; // return optimization not guaranteed
}
```

From personal experience, debug builds will pull a copy of `A` when returning. Move operations can make this cheaper. Release builds don't care, they usually optimize.

bernhardmgruber:
Still, please fix this! :)

PR #238, opened 2022-08-17
Rework example code to illustrate need of move.

hageboeck:
It was discussed that the present example might move and RVO nicely.
Therefore, I reworked it to illustrate that it's hard to pass values
into a function nicely.

I tried to show that pass by copy, pass by const ref and pass by ref all have disadvantages.

Fix https://github.com/hsf-training/cpluspluscourse/issues/106.

### DIFF
--- a/talk/morelanguage/move.tex
+++ b/talk/morelanguage/move.tex
@@ -44,42 +44,67 @@
 
 \begin{frame}[fragile]
   \frametitlecpp[98]{Move semantics: the problem}
-  \begin{exampleblock}{Another potentially non efficient code}
-    \begin{cppcode*}{}
-      MyVector<int> vrandom(unsigned int n) {
-        MyVector<int> result(n);
-        ... // fill result
-        return result;
-      }
-      MyVector<int> v = vrandom(10000);
-    \end{cppcode*}
-  \end{exampleblock}
-  \pause
-  \begin{alertblock}{What could happen on line 4 and 6}
-    \begin{itemize}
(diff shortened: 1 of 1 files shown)


talk/morelanguage/move.tex
```diff
@@ -44,42 +44,67 @@
...
+    \onslide<4>
+    \begin{alertblock}{Use references to avoid copies?}
+      \begin{itemize}
+      \item Working with pass by reference only moves allocation and copy to a different place
+      \end{itemize}
+    \end{alertblock}
+    \onslide<5>
+    \begin{alertblock}{So non-const references?}
+      \begin{itemize}
+      \item Non-\mintinline{cpp}{const} reference could work, but does the outside know that we're changing the vector?
```
bernhardmgruber, line 99:
```suggestion
      \item Non-\mintinline{cpp}{const} references could work, but does the outside know that we're changing the vector?
```

talk/morelanguage/move.tex
```diff
@@ -44,42 +44,67 @@
...
   \end{exampleblock}
-  \pause
-  \begin{block}{The ideal situation}
-    Have a way to express that we move the vector's content
-  \end{block}
+  \begin{overprint}
+    \onslide<2>
+    \begin{alertblock}{Pass by copy}
+      \begin{itemize}
+      \item One unnecessary allocation and one release for a large number of \mintinline{cpp}{int}s
```
sponce, line 86:
```suggestion
      \item One unnecessary (large?) allocation and release
```
Just so that it fits in one line and looks better

sponce (approved):

PR #239, opened 2022-08-17
Added a slide on visitor pattern with lambdas

sponce:
and slightly improved the "Anatomy of a lambda" one to introduce lamda inheritance at that point


talk/morelanguage/lambda.tex
```diff
@@ -344,6 +356,37 @@
   \end{alertblock}
 \end{frame}
 
+\begin{frame}[fragile]
+  \frametitlecpp[17]{lambdas and the visitor pattern}
```
bernhardmgruber, line 360: I wonder whether the slide should be moved to the ones about `std::variant`.

talk/morelanguage/lambda.tex
```diff
@@ -248,40 +248,52 @@
...
+          auto l = __lambda4{sum, off};
+          l(42);
+        \end{cppcode*}
+      \end{column}
+    \end{columns}
+  \end{block}
+  \begin{exampleblock}{Some nice consequence}
+    \begin{itemize}
+    \item lambdas are regular objects
+    \item they can in particular inherit from each other !
```
bernhardmgruber, line 294:
```suggestion
    \item they can in particular be inherited from!
```

talk/morelanguage/lambda.tex
sponce: Well, at that stage we did not yet cover lambdas, so it would be weird.

sponce: issue #246 was open concerning a possible refactoring of the slides there. And we've decided to merge this as is for the moment

PR #248, opened 2022-08-17
Input all tex files directly from the main one

bernhardmgruber:
Refactoring as discussed today.


hageboeck:
LGTM. Eventually, I would prefer having
```
input{1}
input{2]
if not basic input{3}
```
or similar, so it's easier to see what's the difference between the two courses.

(Instead of:)
```
if basic
 1
 2
else
 1
 2
 3
```
but that's for another day.

hageboeck (approved):

bernhardmgruber:
Your proposal is implemented in #223 :)

Issue #246, opened 2022-08-17
Move lambdas before ?exceptions?, use in STL chapter?


hageboeck:
#215

PR #250, opened 2022-08-17
Move lambdas before STL section

bernhardmgruber:
This PR also improves the lambda slides in a few places and moves some slides from the lambda section back into the STL section.

Fixes: #246

- [x] Merge #249 first

### DIFF
--- a/talk/morelanguage/lambda.tex
+++ b/talk/morelanguage/lambda.tex
@@ -65,7 +65,7 @@
     \begin{cppcode*}{firstnumber=4,gobble=2}
       int data[]{1,2,3,4,5};
       auto f = [](int i) {
-        std::cout << i << " squared is " << i*i << std::endl;
+        std::cout << i << " squared is " << i*i << '\n';
       };
       for (int i : data) f(i);
     \end{cppcode*}
@@ -160,7 +160,7 @@
     \begin{cppcode*}{firstnumber=5}
       int sum = 0, offset = 1;
       int data[]{1,9,3,8,3,7,4,6,5};
-      auto f = [&sum, offset](int x) { sum += x + offset; };
+      auto f = [&sum, offset](int x) { sum += x+offset; };
       for (int i : data) f(i);
     \end{cppcode*}
   \end{exampleblock}
--- a/talk/morelanguage/morestl.tex
+++ b/talk/morelanguage/morestl.tex
@@ -176,10 +176,10 @@
     \begin{cppcode*}{gobble=2}
       template<typename... Ts> // covered in expert part
       struct Overload : Ts ... { using Ts::operator()...; };
-      template<typename... Ts> Overload(Ts...) -> Overload<Ts...>;
+      template<typename... Ts> Overload(Ts...) ->Overload<Ts...>;
 
       int main(){
-        std::variant<int, float, std::string, char> v = ...;
+        std::variant<int, float, std::string, char> v{ ... };
         auto overloadSet = Overload {
           [](int i) { std::cout << "i32:" << i;},
           [](float f) { std::cout << "f32:" << f;},
--- a/talk/morelanguage/stl.tex
+++ b/talk/morelanguage/stl.tex
@@ -258,16 +258,16 @@
 \end{frame}
 
 \begin{frame}[fragile]
-  \frametitlecpp[14]{More examples}
+  \frametitlecpp[17]{More examples}
   \begin{cppcode}
     std::list<int> l = ...;
 
     // Finds the first element in a list between 1 and 10.
-    const auto it =
-      std::find_if(l.begin(), l.end(),
+    const auto it = std::find_if(l.begin(), l.end(),
         [](int i) { return i >= 1 && i <= 10; });
-    if (it != l.end())
-      int element = *it;
+    if (it != l.end()) {
+      int element = *it; ...
+    }
 
     // Computes sin(x)/(x + DBL_MIN) for elements of a range.
     std::vector<double> r(l.size());
@@ -280,7 +280,7 @@
 \end{frame}
 
 \begin{frame}[fragile]
-  \frametitlecpp[14]{More examples}
+  \frametitlecpp[11]{More examples}
   \begin{cppcode}
     std::vector<int> v = ...;
 


sponce:
Looks good but for some reason I cannot load the artifacts to check the final result. Anyway, it also needs a rebase, so I'll retry after it

bernhardmgruber:
I rebased now.

sponce:
Reading the slides through, things are not that simple... we use the STL quite a lot in the lamda examples actually !
  - very first example says : 
```cpp
std::vector<int> data{1,2,3,4,5};
std::for_each(begin(data), end(data), [](int i) {
std::cout << "The square of " << i
<< " is " << i*i << std::endl;
});
````
  - then we even use std::transform, non member begin/end and std::variant and std::visit in the last slide.
I fear it has to stay after the STL and even the "more STL"

**Comment by bernhardmgruber** [comment]
The lambda chapter would definitely need more work to strip it of some of the STL content. @hageboeck what do you suggest?

**Comment by hageboeck** [comment]
I prefer having an easy introduction to lambdas, and putting them to good use in the STL chapter later. This way, the syntax is repeated often enough, and people learn how to use lambdas as in e.g.
```c++
STLFunc(iterator, iterator, [](Arg arg) {
  doSomething;
});
```

bernhardmgruber:
@hageboeck I tried to strip the STL part from the lambda chapter, added a few more examples at the end of the STL chapter and moved a few slides.

talk/morelanguage/lambda.tex
```diff
@@ -151,21 +149,19 @@
...
     One can of course mix values and references
     \begin{cppcode*}{firstnumber=5}
       int sum = 0, offset = 1;
-      std::vector<int> data{1,9,3,8,3,7,4,6,5};
-      std::for_each(begin(data), end(data),
-                   [&sum, offset](int x) {
-                     sum += x + offset;
-                   });
+      int data[]{1,9,3,8,3,7,4,6,5};
+      auto f = [&sum, offset](int x) { sum += x + offset; };
```
hageboeck, line 163: This leaks out of the box. Maybe:
```suggestion
      auto f = [&sum, offset](int x) { sum += x+offset; };
```

talk/morelanguage/lambda.tex
```diff
@@ -86,9 +86,9 @@
   \begin{block}{First attempt in \cpp}
```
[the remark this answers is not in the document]
hageboeck: And on a different note:
I would like to see the cppinsights as early as possible to (hopefully) demystify lambdas. What do you think?

hageboeck (commented):
Here my comments for the lambda part. I'll go to STL next.

talk/morelanguage/morestl.tex
```diff
@@ -148,22 +148,50 @@ \subsection{More STL}
...
+    \begin{itemize}
+    \item use inheritance to group a set of lambdas
+    \end{itemize}
+  \end{exampleblock}
+  \begin{block}{Practically}
+    \small
+    \begin{cppcode*}{gobble=2}
+      template<typename... Ts> // covered in expert part
+      struct Overload : Ts ... { using Ts::operator()...; };
+      template<typename... Ts> Overload(Ts...) -> Overload<Ts...>;
```
hageboeck, line 179: This leaks, but just a tiny bit. Maybe it's OK to leave it.

talk/morelanguage/morestl.tex
```diff
@@ -148,22 +148,50 @@ \subsection{More STL}
...
+
+      int main(){
+        std::variant<int, float, std::string, char> v = ...;
```
hageboeck, line 182:
```suggestion
        std::variant<int, float, std::string, char> v{ ... };
```

talk/morelanguage/stl.tex
```diff
@@ -239,18 +258,44 @@
...
-    // Finds the first element in a list between 1 and 10.
     std::list<int> l = ...;
-    ...
+
+    // Finds the first element in a list between 1 and 10.
     const auto it =
       std::find_if(l.begin(), l.end(),
         [](int i) { return i >= 1 && i <= 10; });
+    if (it != l.end())
+      int element = *it;
```
hageboeck, line 270: This is a bit pointless, as `element` will immediately go out of scope.

hageboeck, line 270:
```suggestion
    if (it != l.end()) return *it;
```

hageboeck (commented):
And here the STL part.

talk/morelanguage/lambda.tex
```diff
@@ -113,11 +113,9 @@
   \begin{exampleblock}{Example}
```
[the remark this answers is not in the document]
bernhardmgruber: This is not entirely true, you don't need to capture variables with static storage duration (e.g. global, `static`, some `constexpr`, etc.)

talk/morelanguage/morestl.tex
bernhardmgruber: I know it leaks and I hoped nobody would notice :P I can remove one of the spaces.

talk/morelanguage/stl.tex
bernhardmgruber: I wanted to have the `int element` so you can see that dereferencing the iterator returns an `int`.

bernhardmgruber: Fixed.

talk/morelanguage/lambda.tex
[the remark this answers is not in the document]
bernhardmgruber: I am also not particularly fond of the python code, but it was there when I started working on the course and last time I taught it, one of the students wrote in the feedback that they enjoyed the python analogy made. So I think I would keep it.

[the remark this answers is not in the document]
bernhardmgruber: > I would like to see the cppinsights as early as possible to (hopefully) demystify lambdas. What do you think?

Hmm, I am actually fine with where we demystify lambdas. I think students should first see how lambdas look like and how they are used and then see later how the are implemented.

talk/morelanguage/lambda.tex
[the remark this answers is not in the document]
hageboeck: Indeed. I always captured them without questioning. 🙂 👍

talk/morelanguage/lambda.tex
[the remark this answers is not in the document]
hageboeck: I'm fine with the first Python analogy. That's indeed helpful for people who know it already. But the second one about capture only shows that there's nothing specific to be done in Python. Here I don't see how it helps, because the message is:
In C++ it's different than in Python.

hageboeck (approved):
Very good.

Please consider whether I made a valid point about the 2nd Python lambda (with captures), and then good to merge for me.

PR #252, opened 2022-08-22
Improve lambda slides

bernhardmgruber:
Pulled out from #250.

### DIFF
--- a/talk/morelanguage/stl.tex
+++ b/talk/morelanguage/stl.tex
@@ -199,10 +199,10 @@
           return value + m_inc;
         }
       };
-      std::vector<int> v;
-      v.push_back(5); v.push_back(3); ...
+      std::vector<int> v{1, 2, 3};
+      const auto inc = 42;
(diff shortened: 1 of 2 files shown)


talk/morelanguage/stl.tex
```diff
@@ -199,10 +199,10 @@
           return value + m_inc;
         }
       };
-      std::vector<int> v;
-      v.push_back(5); v.push_back(3); ...
+      std::vector<int> v{1, 2, 3};
+      const auto inc = 42;
```
hageboeck, line 203: Why not use 42 directly? I mean, you can merge as is, but I'm just curious.

hageboeck (approved):
LGTM. Only issue is that all these might be rewritten to introduce lambdas first and STL later.

talk/morelanguage/stl.tex
bernhardmgruber: Because the corresponding lambda example also has a separate variable now to show a capture.

PR #258, opened 2022-09-06
Add credits for the coroutine graph

sponce:


bernhardmgruber (changes requested):
Please fix the spelling errors reported by the pre-commit hook. You can also install it locally, so it does not let you commit without fixing the spelling errors. For this you need to install `pre-commit` and run `pre-commit install` in the repository.

sponce:
I've now setup pre-commit properly. I thought it was already, but somehow was not. Should be fine from now

bernhardmgruber:
I am trying to address the failing pre-commit.ci here: #260.

sponce:
I do not get one thing : how did the codespell succeed for me locally ???

bernhardmgruber:
I think that i may have only checked your changes and those were fine. The spell mistakes were already introduced earlier. E.g. there are spelling issues in ranges, which are outside this PR. Maybe it was also causes by the recent update of codespell/pre-commit.

Issue #78, opened 2021-12-15
Introduce a section on best practices


theanalyst:
Since this is a fairly generic idea, for starts we could introduce a simple slide linking to Core Guidelines (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines). Maybe we could cherry-pick a few guidelines that are commonly considered good, for eg:
- Interfaces -I.3 (singletons) as well as I.11 & I.13 (avoid raw_ptr ownership transfer, avoid passing arrays without length)
- Errors - E.6 use RAII
- Classes - C.131 Avoid trivial getters/setters 

(Fairly opinionated list ;)

hageboeck:
Some more ideas:
- [ ] If you have a debate about how to do something, check core guidelines
- [ ] When to use free functions (anonymous namespace) and when member functions
- [ ] Keep headers clean, low number of includes?

And on the repetition side:
- [ ] Use smart pointers
- [ ] Pass by reference

theanalyst:
> Some more ideas:
> 
>     * [ ]  If you have a debate about how to do something, check core guidelines
Definitely. I feel like for new comers this should give some pointers on doing things in mostly correct way 
>     * [ ]  When to use free functions (anonymous namespace) and when member functions
> 
>     * [ ]  Keep headers clean, low number of includes?
:+1:

> 
> 
> And on the repetition side:
> 
>     * [ ]  Use smart pointers
>     * [ ]  Pass by reference

Yeah this would be nice, esp. understanding ownership, ie. naked pointers as a view only resource

krasznaa:
As we discussed just now in our meeting, here are the ATLAS coding standards. Which could be pointed to, together with those of the other experiments. :wink:

http://atlas-computing.web.cern.ch/atlas-computing/projects/qa/draft_guidelines-0.2.html

hageboeck:
Decided to make a box/design for making good practices easily recognisable, and put the good practices right next to the places where the rationales can be explained.

bernhardmgruber:
I think we should discuss the SOLID principles at the end of the OOP chapter to give some guidance: https://en.wikipedia.org/wiki/SOLID

hageboeck:
Closing in favour of #507

PR #259, opened 2022-09-06
Good practice

hageboeck:
To continue on #78, here's a draft of a recognisable good practice box. What's missing is the "index of good practices". I need to learn a bit more on the deeper tex mechanics to do this cleanly.
The new environment will in the future "just" create that index automatically.

### DIFF
--- a/talk/setup.tex
+++ b/talk/setup.tex
@@ -31,6 +31,9 @@
   \end{beamercolorbox}%
 }
 
+\setbeamercolor{goodpractice head}{fg=white,bg=orange!85!black}
+\setbeamercolor{goodpractice body}{fg=black,bg=orange!25}
+
 %%%%%%%%%%%%
 % packages %
 %%%%%%%%%%%%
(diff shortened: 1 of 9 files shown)


talk/setup.tex
```diff
@@ -31,6 +31,9 @@
   \end{beamercolorbox}%
 }
 
+\setbeamercolor{goodpractice head}{fg=white,bg=orange}
```
sponce, line 34:
```suggestion
\setbeamercolor{goodpractice head}{fg=white,bg=orange!50!black}
```
Or something of this kind, that is more dark so that the white writing is easier to read

sponce (approved):

bernhardmgruber (approved):

talk/setup.tex
hageboeck: `orange!50!black`:
image

`orange!90!black`:
image


Current head:
image

Opinions?

sponce: some 70-80 ? so that it stays orange (and not brown) but slightly darker. Actually the 90 is not bad already

hageboeck: 85 now. I will merge, and we can tweak it once we see the slides on a projector.

PR #260, opened 2022-09-06
Fix spelling errors reported by codespell

bernhardmgruber:


codespell.txt
```diff
@@ -3,3 +3,4 @@ heist
 Hist
 Heist
 his
+ro
```
bernhardmgruber, line 6: So codespell flagged the `ro` (return object varibale) in the new coroutine flowchart. I could not figure out how to silence codespell on just one line, so I added ro to the exceptions here. Does anyone know a better way?

sponce: In my experience, automatic codespell is bound to fail. So we can keep it as an hint, but we'll have to ignore it most of the time. Never seen a better way

bernhardmgruber: What do you mean with "keep it as an hint"? Should I add a comment in the codespell.txt why I added `ro` to the exceptions?

sponce: I mean keep the codespell so that we can go through its output from time to time and fix what needs to be fixed. But ignore things like "ro". By "hint", I meant something that does not constrain us but helps finding out problems.

Anyway, for now, let's merge this

bernhardmgruber: Hmm. But having a CI job that we are ignoring anyway, because it's always red, is also not a good practice :) I would prefer if we can keep it green.

sponce: everybody would, but you've just demonstrated it will be tough. And this is my experience indeed with spell checking

Issue #267, opened 2022-09-21
Add index of exercises on slides


bernhardmgruber:
@hageboeck and me discussed that it would be useful to have an index of all exercises at the end of the slides.

Issue #245, opened 2022-08-17
Advanced course contents


hageboeck:
| Day | Contents |
|----|-----------|

sponce:
So let's try to have a first go at the content of an advanced course, chapter by chapter.
I propose to not do a dedicated pdf so that in case we need to go back to more basic concepts than expected, we have the whole set of slides. We could however create a second outline slide dedicated to advanced course, so that we know what to cover by simply going back to it.

Language basics -> as a short intro, a test of what people know and a recap of C++11/14/17 new features
--
  - scopesnamespaces (only unnamed namespace)
  - classenum (only enum class + using)
  - references (1 slide)
  - control (only init statements for if and switch & range based loops)
  - inline
  - assert

Object orientation (OO)
--
  - typecasting
  - operators (insisting on non member and friends)
  - functors
  - adl

Core modern \cpp
--
  - constexpr (consteval and constinit)
  - exceptions (noexcept and alternatives)
  - move (with no intro on the problem)
  - copyelision
  - templates (quick recap and CTAD)
  - lambda (quick recap, this and generic lambdas)
  - morestl
  - ranges
  - raii
  - random
  - initialization

Expert \cpp
--
  - variadictemplate
  - perfectforwarding
  - sfinae
  - cpp20concepts
  - cpp20spaceship
  - modules
  - coroutines

Useful tools
--
  - formatting
  - webtools
  - sanitizers
  - profiling
  - doxygen

Concurrency
--
  - threadsasync
  - mutexes
  - atomic
  - condition

Concerning the timing and order : 
  - let's start the the basics and OO part as a recap
  - half of core modern C++ down to move would probably also fit the first day
  - rest of core modern + tools and concurrency on 2nd day
  - as much as we can from expert on last day, providing we may not cover everything there. I would do the first 2, then spaceship, sfinae and concepts and that's probably it.

To be tested live obviously

hageboeck:
I think it's a good curriculum. The only complication is:
Once you finish a topic, you cannot go to the next slide, but you have to go back to the index. It's harder on the speaker.

Can we try to have this 2nd index slide by the next meeting? My (so far) unsuccessful attempts at creating an index of good practices makes me think that this will be more work than we might wish for.

sponce:
For good practices, it should not be so difficult. Just follow the example here : https://tex.stackexchange.com/questions/564098/how-to-add-include-the-title-of-custom-new-list-of-on-the-toc

[...]

PR #268, opened 2022-09-22
Create an index of good practices

hageboeck:
Augment the goodpractice environment with automatic entries in the list
of good practices (C++Course.lgp).

The index shows up at the end of the course.

image

The same technique could be used to generate an index of all exercises (#267).

Regarding @sponce 's comment in #245:
> For good practices, it should not be so difficult. Just follow the example here:
> https://tex.stackexchange.com/questions/564098/how-to-add-include-the-title-of-custom-new-list-of-on-the-toc

> You basically create a new kind of list, here listing the good practices (maybe call it \listgoodpractice and associate a new command \goodpractice which adds to the list and another one \listofgoodpractices which displays the lit, using the \tocfile command form tocloft package.
Then you only have to call \goodpractice in the definition of your goodpractice block

tocloft doesn't work with beamer. But I guess the stuff I'm doing with `\addtocontents` is simple enough for it not to break in the future.

### DIFF
--- a/talk/C++Course.tex
+++ b/talk/C++Course.tex
@@ -201,4 +201,11 @@
   \end{center}
 \end{frame}
 
+\section{Index}
+
+\begin{frame}
+  \frametitle{Index of Good Practices}
+  \listofgoodpractices
+\end{frame}
+
 \end{document}
--- a/talk/C++Course.tex
+++ b/talk/C++Course.tex
@@ -201,7 +201,7 @@
   \end{center}
 \end{frame}
 
-\section{Index}
+\section*{Index}
 
 \begin{frame}
   \frametitle{Index of Good Practices}
--- a/talk/Makefile
+++ b/talk/Makefile
@@ -13,7 +13,7 @@ endif
 	pdflatex -shell-escape C++Course.tex
 
 clean:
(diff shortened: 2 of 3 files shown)


hageboeck:
There something I need to figure out still:
image

bernhardmgruber:
I don't think I can help you reviewing the .tex changes :D

hageboeck:
> I don't think I can help you reviewing the .tex changes :D

I'm happy if you check the slides that drop out of the CI.

sponce (approved):

bernhardmgruber:
I think there is something wrong with the slide number:
image
But then:
image

hageboeck:
> I think there is something wrong with the slide number:

Indeed, they are page numbers in the PDF. I can change that into the slide number, though. The link will still go to the page, but it will display the slide number.
Sounds good, @bernhardmgruber ?

bernhardmgruber:
> > I think there is something wrong with the slide number:
> 
> Indeed, they are page numbers in the PDF. I can change that into the slide number, though. The link will still go to the page, but it will display the slide number. Sounds good, @bernhardmgruber ?

Oh, I did not notice :D Maybe show both?

FWIW, in Adobe Reader I also enter the slide number when jumping around, not the PDF page number:
image

Issue #263, opened 2022-09-21
Create example for variadic templates


roiser:
... placeholder

PR #282, opened 2022-09-23
Add exercise on variadic templates

bernhardmgruber:
Fixes: #263

### DIFF
--- a/code/variadic/README.md
+++ b/code/variadic/README.md
@@ -1,0 +1,22 @@
+
+## Instructions
+
+In this exercise, we will take the `tuple` implementation from the slides
+and extend it step by step to make the example code in `main` compile.
+
+Feel free to comment out parts of main to test parts of your implementation sooner.
+All instructions are given in more detail in the source code as well, especially inside `main`.
+
+This example neglects const correctness so there is less code to write.
+Making it const correct would require adding several overloads with almost the same code as the non-const version.
+You can try this is a bonus exercise at the end!
+
+1. make `tuple` default-constructible
+2. avoid implicit construction from a single value
+3. write `head(tuple)`, which returns the first element of a tuple
+4. implement `get<I>(tuple)`
+5. specialize `std::tuple_size`
+6. specialize `std::tuple_element`
+7. implement a binary `tuple_cat(tuple, tuple)`
+8. implement `sum(tuple)`
+9. implement `operator<<`, producing comma separated output
--- a/code/variadic/solution/variadic.sol.cpp
+++ b/code/variadic/solution/variadic.sol.cpp
@@ -1,0 +1,183 @@
+#include <cassert>
+#include <iomanip>
+#include <iostream>
+#include <sstream>
+#include <tuple>
+
+// we start with the tuple from the slides on variadic templates
+
+template <typename... Ts>
+struct tuple {};
+
+template <typename T, typename... Ts>
+struct tuple<T, Ts...> : tuple<Ts...> {
+    // exercise 1: activate the default constructor
+    tuple() = default;
+
+    // exercise 2: make constructor explicit
+    explicit tuple(T head, Ts... tail) : tuple<Ts...>(tail...), m_head(head) {}
+
+    T m_head;
+};
+
+// we need a deduction guide, because the compiler only searches the primary template for constructors
+template <typename... Ts>
+tuple(Ts...) -> tuple<Ts...>;
+
+// here is a helper to get the tail of a tuple
+template <typename Head, typename... Tail>
+auto& tail(tuple<Head, Tail...>& t) {
+    // we just cast the tuple to its base class, which holds 1 element less
+    return static_cast<tuple<Tail...>&>(t);
+}
+
+// exercise 3: implement head(tuple) here
+template <typename... Ts>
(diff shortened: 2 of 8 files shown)


code/variadic/README.md
```diff
@@ -0,0 +1,18 @@
...
+Feel free to comment out parts of main to test parts of your implementation sooner.
+All instructions are given in more detail in the source code as well.
+
+1. make `tuple` default construct
+2. avoid implicit construction from a single value
+3. write `head(tuple)`, which returns the first element of a tuple
+4. implement `get<I>(tuple)`
+5. specialize `std::tuple_size`
+6. specialize `std::tuple_element`
+7. implement `operator<<` for `tuple`
```
sponce, line 16:
```suggestion
7. implement simple `operator<<` for `tuple`
```

code/variadic/README.md
```diff
@@ -0,0 +1,18 @@
...
+8. implement a binary `tuple_cat(tuple, tuple)`
+9. make `operator<<` produce a given output
```
sponce, line 18:
```suggestion
9. make complete `operator<<` producing comma separated output
```

code/variadic/solution/variadic.sol.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include <tuple>
+
+// we start with the tuple from the slides on variadic templates
+
+template <typename... Ts>
+struct tuple {};
+
+template <typename T, typename... Ts>
+struct tuple<T, Ts...> : tuple<Ts...> {
+    tuple() = default;
```
sponce, line 14:
```suggestion
    // exercise 1: activate the default constructor
    tuple() = default;
```

code/variadic/solution/variadic.sol.cpp
```diff
@@ -0,0 +1,167 @@
...
+template <typename... Ts>
+auto& head(tuple<Ts...>& t) {
+    return t.m_head;
+}
+
+// exercise 4: implement get<I> here
+template <std::size_t I, typename... Ts>
+auto& get(tuple<Ts...>& t) {
+    if constexpr (I == 0)
+        return t.m_head;
```
sponce, line 42:
```suggestion
        return head(t);
```

bernhardmgruber:
> I would simply drop exercise 7. The fold is not really trivial (and actually throws a wrong compiler warning to me that the statement has no effect ) and anyway the output is crap.

I think the fold `(os << ... << get<Is>(t));` is easier than `((os << get<Is>(t) << (Is == sizeof...(Ts) - 1 ? "" : ", ")), ...);`. And the first fold would be an example of a binary fold. Exercise 9 uses a unary fold over comma, which always feels like an abuse to me.

The reason you may get a compiler warning is because printing the empty tuple let's the fold collapse to just `os;`, which is indeed a statement without an effect.

bernhardmgruber:
I could add another exercise where the students are supposed to implement a function `add(tuple<...> t)` that just does a fold over `+` and adds the tuple elements? That could be an example for a binary fold, since it should handle an empty tuple. What do you think?

sponce:
> I could add another exercise where the students are supposed to implement a function `add(tuple<...> t)` that just does a fold over `+` and adds the tuple elements? That could be an example for a binary fold, since it should handle an empty tuple. What do you think?

sounds like a good idea indeed if you want to show the binary fold. At least I find it better than implementing a "bad" operator<< (in the sens that the output is unreadable without spaces or commas)

bernhardmgruber:
I added an addition example to implement `auto sum(tuple<Ts...>)`.

code/variadic/README.md
```diff
@@ -0,0 +1,22 @@
...
+and extend it step by step to make the example code in `main` compile.
+
+Feel free to comment out parts of main to test parts of your implementation sooner.
+All instructions are given in more detail in the source code as well.
+
+This example neglects const correctness so there is less code to write.
+Making it const correct would require adding several overloads with almost the same code as the non-const version.
+You can try this is a bonus exercise at the end!
+
+1. make `tuple` default construct
```
hageboeck, line 14: For me this doesn't parse.
- *a* default construct?
- default construct*able*
- Ensure that `tuple` has a default constructor

By looking at the solution, the best is presumably:
```suggestion
1. Ask the compiler to generate a default constructor for `tuple`
```

hageboeck: I would say that the instructions should be much more detailed. I struggle to understand what to do. More suggestions to follow soon

bernhardmgruber: And the statement means: "Make tuple<T...> t; compile". So, make `tuple` default constructible would be the best of your wordings.

bernhardmgruber: Yes, because this is where they should add some code as detailed inside `main` :)

hageboeck: Ah, OK. I didn't manage to read that far down before the meeting. Now continuing ...

bernhardmgruber: I rephrased `make tuple default construct` to `make tuple default-constructible` and pointed out more strongly in the README.md that the detailed instructions are to be found in `main`.

sponce (approved):

PR #284, opened 2022-09-24
Use different structure color on advanced slides

bernhardmgruber:
Based on PR #223, changing the formatting of advanced slides is now (almost) trivial. This PR proposes to color advanced slides violet instead of blue.

- [x] Merge #223 first

### DIFF
--- a/talk/C++Course.tex
+++ b/talk/C++Course.tex
@@ -12,9 +12,14 @@
 
 % create a comment environment advanced. depending on the value of basic, we exclude or include it.
 \ifbasic
-\excludecomment{advanced}
+  \excludecomment{advanced}
 \else
-\includecomment{advanced}
+  % advanced slides have a different structure color
+  \specialcomment{advanced}{
+    \setbeamercolor{structure}{fg=beamer@blendedblue!60!violet}
+  }{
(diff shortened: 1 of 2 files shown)


bernhardmgruber:
Random basic slide:
image

Random advanced slide:
image

sponce (commented):
Works nicely but I must say the the violet hurts my eyes... I liked the original colors, they are not the default in beamer just by chance. It's the most readable. So I'm not so keen on changing them

sponce:
What about this color ?
slide

bernhardmgruber:
> I've just replace violet by beamer@blendedblue!70!violet

I think the difference between basic and advanced slides is too little. It might not even be visible on bad projectors.

sponce:
It's precisely done so that it does not seem too different but is clearly if you look for it. Maybe put 60 instead of 70 then, but clearly not 100% violet.

bernhardmgruber:
> It's precisely done so that it does not seem too different but is clearly if you look for it. Maybe put 60 instead of 70 then, but clearly not 100% violet.

I tried `beamer@blendedblue!60!violet` and it is still difficult to see the difference. When I make the slides fullscreen and browse through them quickly, I can notice the difference though. I would prefer a stronger distinction, but let's go with it for now!

PR #285, opened 2022-09-28
Improved solution and instructions of operator exercise

### LINKED ISSUE #264: revisit operartors exercise

sponce:
Insist on non member operators and use friend (only) when needed

Closes #264

### DIFF
--- a/code/operators/README.md
+++ b/code/operators/README.md
@@ -2,4 +2,5 @@
 ## Instructions
 
 * inspect main and complete the implementation of class Fraction step by step
-* you can comment out parts of main to test in between
+  * you can comment out parts of main to test in between
+* when possible, prefer non-member operators and friend for operators
--- a/code/operators/solution/operators.sol.cpp
+++ b/code/operators/solution/operators.sol.cpp
@@ -6,88 +6,77 @@
 class Fraction {
  public:
   explicit Fraction(int i) : m_num(i), m_denom(1) {}
-  Fraction(int num, int denom) : m_num(num), m_denom(denom) {}
+  Fraction(int num, int denom) : m_num(num), m_denom(denom) {
+    normalize();
+  }
 
-  int num() const { return m_num; }
-  int denom() const { return m_denom; }
+  friend std::ostream& operator<<(std::ostream& os, Fraction const & f) {
+    os << f.m_num << "/" << f.m_denom;
+    return os;
+  }
 
+  Fraction & operator*=(int i) {
+    m_num *= i;
+    normalize();
+    return *this;
+  }
+
+  Fraction & operator*=(Fraction const & o) {
+    m_num *= o.m_num;
+    m_denom *= o.m_denom;
+    normalize();
+    return *this;
+  }
+
+  friend Fraction operator*(Fraction f, int i) { return f *= i; }
+  friend Fraction operator*(int i, Fraction const & f) { return f * i; }
+  friend Fraction operator*(Fraction a, Fraction const & b) { return a *= b; }
+
+  Fraction & operator+=(int i) {
+    m_num += i * m_denom;
+    return *this;
+  }
+
+  Fraction & operator+=(Fraction o) {
+    m_num *= o.m_denom;
+    m_num += o.m_num * m_denom;
+    m_denom *= o.m_denom;
+    normalize();
+    return *this;
+  }
+
+  friend Fraction operator+(Fraction r, int i) { return r += i; }
+  friend Fraction operator+(int i, Fraction const & r) { return r + i; }
+  friend Fraction operator+(Fraction a, Fraction const & b) { return a += b; }
+
+  friend bool operator==(Fraction const & a, Fraction const & b) {
+    return a.m_num == b.m_num && a.m_denom == b.m_denom;
+  }
+
+  friend bool operator<(Fraction const & a, Fraction const & b) {
+    return a.m_num * b.m_denom < b.m_num * a.m_denom;
+  }
+
+  friend bool operator!=(Fraction const & a, Fraction const & b) { return !(a == b); }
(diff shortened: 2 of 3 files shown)


code/operators/README.md
```diff
@@ -2,4 +2,5 @@
 ## Instructions
 
 * inspect main and complete the implementation of class Fraction step by step
-* you can comment out parts of main to test in between
+  * you can comment out parts of main to test in between
+* try to use as much a possible non member operators and friedn when needed
```
bernhardmgruber, line 6:
```suggestion
* try to use as much a possible non-member operators and friend when needed
```
Or:
```suggestion
* when possible, prefer non-member operators and friend for operators
```

code/operators/solution/operators.sol.cpp
```diff
@@ -6,88 +6,77 @@
...
-Fraction operator*(int i, Fraction r) { return r * i; }
-
-Fraction operator*(Fraction a, Fraction b) { return a *= b; }
+Fraction operator*(Fraction f, int i) { return f *= i; }
+Fraction operator*(int i, Fraction const & f) { return f * i; }
+Fraction operator*(Fraction a, Fraction const & b) { return a *= b; }
 
 Fraction operator+(Fraction r, int i) { return r += i; }
+Fraction operator+(int i, Fraction const & r) { return r + i; }
+Fraction operator+(Fraction a, Fraction const & b) { return a += b; }
```
bernhardmgruber, line 68: These should be hidden friends.

code/operators/solution/operators.sol.cpp
```diff
@@ -6,88 +6,77 @@
...
-bool operator>(Fraction a, Fraction b) { return b < a; }
-
-bool operator<=(Fraction a, Fraction b) { return !(a > b); }
-
-bool operator>=(Fraction a, Fraction b) { return !(a < b); }
-
+bool operator!=(Fraction const & a, Fraction const & b) { return !(a == b); }
+bool operator>(Fraction const & a, Fraction const & b) { return b < a; }
+bool operator<=(Fraction const & a, Fraction const & b) { return !(a > b); }
+bool operator>=(Fraction const & a, Fraction const & b) { return !(a < b); }
```
bernhardmgruber, line 73: More hidden friends here.

code/operators/solution/operators.sol.cpp
```diff
@@ -6,88 +6,77 @@
...
 
-  void normalize() {
-    const int gcd = std::gcd(m_num, m_denom);
-    m_num /= gcd;
-    m_denom /= gcd;
+  friend Fraction & operator*=(Fraction & f, int i) {
+    f.m_num *= i;
+    f.normalize();
+    return f;
   }
```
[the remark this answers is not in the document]
sponce: And why not ? Why would you make a difference with e.g. the operator* ?

code/operators/solution/operators.sol.cpp
bernhardmgruber: Here you define operators as free function. This adds them visibly to the global namespace. So, e.g. whenever you add two `std::string` like: `std::string a, b; a + b;`, the compiler will have to try your operators from `Fraction`  as well (and discard them). This causes a compile-time slowdown, because the overload set for `operator+` is unnecessary large. Here is a blog post with a synthetic benchmark claiming a 23x (!) reduction in compilation time using hidden friends: https://jacquesheunis.com/post/hidden-friend-compilation/

Secondly, if you now defined a class `Complex`, that had an implicit conversion operator to `Fraction`, your `Complex` would use all the operators from `Fraction`, because the compiler would just figure out the right conversion sequence. So a call to `Complex a, b; a + b` would be translated to: `Complex a, b; operator+()(Fraction(a), Fraction(b))`, which does not sound too crazy now, but can actually lead to big surprises. Again, hidden friends would not allow such a conversion because ADL could not find the `operator+` inside `Fraction` for two `Complex` arguments. More in this blob post: https://www.justsoftwaresolutions.co.uk/cplusplus/hidden-friends.html

code/operators/solution/operators.sol.cpp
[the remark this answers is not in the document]
bernhardmgruber: Here, the operator is naturally asymmetric, so `a *= b` is expected to behave differently than `b *= a`. So there is no point in making it a free function for the sake of symmetry (like with `operator*()(T, T)` for example. Furthermore, if it is an ordinary member function (which it was before this PR), then it is also not visible outside the class `Fraction` (and does not pollute the namespace) and no other class with a conversion operator to `Fraction` could hijack it.

code/operators/solution/operators.sol.cpp
```diff
@@ -6,88 +6,77 @@
...
+  Fraction & operator*=(int i) {
     m_num *= i;
     normalize();
     return *this;
   }
 
-  Fraction& operator*=(Fraction f) {
-    m_num *= f.num();
-    m_denom *= f.denom();
+  Fraction & operator*=(Fraction const & o) {
```
bernhardmgruber, line 24: It hurts my soul to see you trample my beautiful code with these unnecessary whitespaces and use of east-const :)

Should we add a `.clang-format` at some point to format the examples? Since clang-format-14 it can also handle east/west const.

bernhardmgruber (approved):

bernhardmgruber:
Btw, in reaction your our discussion here I extended the slides on hidden friends and gave guidance for operator overloading #288.

code/operators/solution/operators.sol.cpp
sponce: east const is the only logical way to use const keyword. west const should just be banned in my opinion.
But yes, I agree, we should have a .clang-format at some point. Not the most urgent though.

PR #288, opened 2022-09-29
Extend slides on hidden friends and give guidance on operator overloading

bernhardmgruber:


talk/objectorientation/adl.tex
```diff
@@ -168,3 +173,62 @@
...
+      \item Are \emph{only} found via ADL, e.g.\ \mintinline{cpp}{std::cout << complex;}
+      \item Compiler needs to consider less functions during ordinary name lookup (faster compilation)
+      \item Avoid accidental implicit conversions
+    \end{itemize}
+  \end{block}
+  \begin{alertblock}{Accidental conversion example}
+    \small
+    \begin{cppcode*}{gobble=2}
+      std::ostream & operator<<(std::ostream & os,
+        Complex const & c) { ... } // out-of-class definition
```
sponce, line 190:
```suggestion
      std::ostream & operator<< // out-of-class definition
        (std::ostream & os, Complex const & c) { ... }
```

talk/objectorientation/adl.tex
```diff
@@ -168,3 +173,62 @@
...
+      // either: Complex::Complex(Fraction f) { ... }
```
sponce, line 191: Why not put it as uncommented code inside the struct ?

talk/objectorientation/adl.tex
```diff
@@ -168,3 +173,62 @@
...
+      struct Fraction {
+        int num, denom;
+        operator Complex() const { ... } // or: conversion op
+      };
+      std::cout << Fraction{2, 3}; // Prints 0.6666 + 0i
+      // operator<<(std::cout, Complex(Fraction{2, 3}));
+      // operator<<(std::cout, Fraction{2, 3}.operator Complex());
```
sponce, line 198: Why are these lines commented ? They would compile. So strange things, but compile

sponce (approved):

talk/objectorientation/adl.tex
bernhardmgruber: Because this is a constructor for `Complex`, which is not defined on this slide.

talk/objectorientation/adl.tex
bernhardmgruber: Because these lines show what would we called in the two cases. I will make that more clear.

PR #290, opened 2022-09-30
Add a slide on index_sequence with lambda

bernhardmgruber:
We should also show how to use `std::index_sequence` with a lambda, since this use should become more comment as C++20 becomes available.


hageboeck:
Well, maybe it's better to declare the lambda first before calling it. That hopefully improves the readability?

bernhardmgruber:
> Well, maybe it's better to declare the lambda first before calling it. That hopefully improves the readability?

I chose the IILE (immediately invoked lambda expression) exactly because it is the style as shown in various conference talks and because it fulfills the IILE pattern :)

Issue #244, opened 2022-08-17
Exercise on concepts


hageboeck:
We would like to see an exercise on concepts

hageboeck:
The exercise is already there; closing now.

PR #291, opened 2022-10-02
Create index of exercises

hageboeck:
Replicate the work for the `goodpractice` environment to create a list of exercises. The current list is:

image

**NB**: The "Exercise time" on the left is already fixed.

Do you know of an exercise that I might have missed?
@bernhardmgruber @sponce

### DIFF
--- a/talk/C++Course.tex
+++ b/talk/C++Course.tex
@@ -206,4 +206,9 @@
   \listofgoodpractices
 \end{frame}
 
+\begin{frame}
+  \frametitle{Index of Exercises}
+  \listofexercises
+\end{frame}
+
 \end{document}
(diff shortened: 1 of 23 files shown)


bernhardmgruber:
For some reason, the templates exercise still appear as "Exercise time" in your table of exercises.

bernhardmgruber:
Some of the exercise names start lowercase. I wonder whether we should make them uppercase. At least the ones which are not literal names of tools. So at least `Smart pointers` :)

bernhardmgruber:
The modules exercises are missing. But because I have not added an exercise box for them :D Will fix that!

bernhardmgruber:
Address sanitizer exercise is missing.

bernhardmgruber:
Oh yeah, we have two outstanding exercises in #282 and #244.

hageboeck:
> For some reason, the templates exercise still appear as "Exercise time" in your table of exercises.

Usually, you need to compile twice, because in the first pass, the index gets recreated, and typeset in the second. I have a version where it shows up correctly.

hageboeck:
I'll merge like this to unblock other PRs. Suggestions for improving this are welcome as additional PRs!

PR #295, opened 2022-10-02
Various small improvements

bernhardmgruber:
I collected these while I was browsing through the slides to collect if all exercises were collected in #291.

### DIFF
--- a/talk/expert/sfinae.tex
+++ b/talk/expert/sfinae.tex
@@ -173,7 +173,7 @@
       struct A{}; struct B{ void foo(); };
       struct C{ int foo(); };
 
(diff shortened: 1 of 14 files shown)


talk/expert/sfinae.tex
```diff
@@ -268,13 +268,13 @@
 \begin{frame}[fragile]
```
hageboeck, line 268: I'm not allowed to comment above, but in line 252 and 258, `enable_if_t` is not c++11.

hageboeck (approved):
Very good. If you have the time, there's a c++11 that should become c++14.

hageboeck:
👍 for me, you can merge now.

PR #297, opened 2022-10-03
Try macos-12 for CI

hageboeck:
To allow compiling c++20 and concepts, try out macos12.

Attempt to fix CI for #261.


hageboeck:
@krasznaa `macos-latest` == 11 doesn't work with concepts. Do you see any reason not to move to `macos-12`?

bernhardmgruber (approved):

krasznaa (approved):
Sure, we can switch to this hardcoded version. We'll need to remember to advertise it to the users as well that they'd need the latest OS / Xcode version to make the tutorial work.

PR #298, opened 2022-10-03
Try macos12 with concepts exercise

hageboeck:
This is a combination of #261 and #297.


hageboeck:
Conclusion: #261 with macos12 works.

chavid:
Do I have smething to do on my side, so that macos12 is used ?

PR #299, opened 2022-10-04
Automatic update of exercise count on first page now that we have a counter for it

sponce:
trivial change


hageboeck (approved):
Bravo, good idea.

PR #300, opened 2022-10-04
Moved to C++20 for exercises and checked it's supported in hello exercise

sponce:


bernhardmgruber:
This PR only affects cmake and not all the individual makefiles (except for the hello exercise). is this intended?

sponce:
hum... not sure. I realize now that the same CompilerSettings.cmake in included in all tests. I initially only wanted to modify the hello example. Now, putting C++20 for everyone would allow to avoid overwriting the C++ standard in the C++20 examples, like variadic, concepts and spaceship.
What do you think ? I can just force the standard in hello.

bernhardmgruber:
Well, you either only require C++20 in the hello example (both cmake and Makefile), or we switch ALL examples (both cmake and Makefiles) to C++20. I think I would rather only switch the hello example for now.

PR #310, opened 2022-10-06
Fixed tex compilation

sponce:


chavid:
texttt dislikes :: ?

sponce:
No, but latex takes '_' as a special character (for subscript). So you need `\_`. Unless you use a verbatim like environment, here minted.

chavid:
> No, but latex takes '_' as a special character (for subscript). So you need `\_`. Unless you use a verbatim like environment, here minted.

Got it. Thanks.

PR #311, opened 2022-10-06
Improved first page, with automatic date in particular

sponce:


talk/setup.tex
```diff
@@ -335,9 +335,9 @@
 %%%%%%%%%%%%%%%%%%
 
 \title{HEP \cpp course}
-\author[B. Gruber, S. Hageboeck, S. Ponce]{Based on the work of \\ S\'ebastien Ponce \\ \texttt{sebastien.ponce@cern.ch}}
+\author[B. Gruber, S. Hageboeck, S. Ponce]{B. Gruber, S. Hageboeck, S. Ponce \\ \texttt{sebastien.ponce@cern.ch}}
```
bernhardmgruber, line 338: @sponce you still did the lionshare of all the work over many years. I think it's fine to point that out still somewhere!

At least I feel I should not be named in the same way you are.

sponce: I count on git to prove what was done by whom. And you may have noticed that only my mail is there. Let's go like this.

PR #315, opened 2022-10-06
Add notes to cheat sheet.

hageboeck:
@chavid 

Is this more or less how the tutors should see the exercises?


code/ExercisesCheatSheet.md
```diff
@@ -140,6 +140,19 @@ Here we have four code snippets that will benefit from using smart pointers.
 - `problem3` is an example of shared ownership where `std::shared_pointer` should be used.
 - `problem4` demonstrates the usage of `shared_ptr` as class members. It has a second part where a `weak_ptr` can be used, but can be skipped if not enough time.
 
+### std::optional (directory: `optional`)
+Use std::optional to signify disallowed values in a computation.
+1. Use std::optional as return value of the mysqrt function for negative arguments
```
sponce, line 145:
```suggestion
1. Use std::optional as return value of the mysqrt function. Use `nullopt_t` for negative arguments. Note that `return {}` will create a `nullopt_t` automatically.
```

code/ExercisesCheatSheet.md
```diff
@@ -140,6 +140,19 @@ Here we have four code snippets that will benefit from using smart pointers.
...
+2. Given that the return type changes, modify the square function accordingly
```
sponce, line 146:
```suggestion
2. Given that the return type changes, modify the square function accordingly to take into account cases where a `nullop_t` is given as input. Note that `std::optional` can be directly used in an if statement as if it would be a boolean to check whether is value is present
```

sponce (approved):

Issue #287, opened 2022-09-29
Question on ADL consequence slide


bernhardmgruber:
In the section on ADL, we have this slide:
![image](https://user-images.githubusercontent.com/1224051/193033577-ecc459c2-8817-4fbc-ba99-2f24acede2d5.png)

I don't get what this recommendation has to do with ADL. Can someone explain this to me please?

bernhardmgruber:
@sponce, any idea?

sponce:
Not really. I also do not see the relation with ADL just like this...

bernhardmgruber:
Hmm. I was wondering whether it meant something like: If you want to add functionality to a framework, implement this functionality inside the namespaces of the framework.

If it's about using a framework with namespaces, I don't see a problem with the `using namespace MyProject`.

bernhardmgruber:
I propose to remove the slide: #316

bernhardmgruber:
@hageboeck or @chavid, do you know what this slide was intended to convey?

chavid:
No idea where this comes from.

PR #316, opened 2022-10-07
Remove confusing slide on ADL consequences

bernhardmgruber:
Related to: #287


sponce:
Sounds good to me. Let's drop it. Why id this still draft ?

sponce (approved):

bernhardmgruber:
> Sounds good to me. Let's drop it. Why id this still draft ?

I wanted to leave some time before we merge it, so maybe someone who knows why we have the slide could react. But ok, time's up :)

PR #319, opened 2022-10-07
Mark a few extra slides as 'advanced'

sponce:


talk/basicconcepts/scopesnamespaces.tex
```diff
@@ -155,8 +155,9 @@
   \end{exampleblock}
 \end{frame}
 
+\begin{advanced}
 \begin{frame}[fragile]
-  \frametitlecpp[98]{Unnamed namespaces}
+  \frametitlecpp[98]{Anonymous namespaces}
```
bernhardmgruber, line 160: I think I even changed this at some point to "unnamed namespaces" because this is term used on cppreference: https://en.cppreference.com/w/cpp/language/namespace

"Anonymous namespaces" are a popular alias. Can you maybe mention both terms?

sponce: Interesting, I had never heard of "unnamed namespace". I've put both in the title now :-)

PR #322, opened 2022-10-10
Improvements to concurrency chapter

hageboeck:
Four commits, in order of importance:
- Rework slides on condition_variable.
- Small tweaks in condition_variable exercise.
- Small improvements to concurrency chapter.
- [NFC] Use {} for init and \n for endl in concurrency chapter.

### DIFF
--- a/code/condition_variable/condition_variable.cpp
+++ b/code/condition_variable/condition_variable.cpp
@@ -13,13 +13,12 @@
  */
 using namespace std::chrono_literals; // We can write 1s
 
-std::mutex cout_mutex; // We need this to synchronise printing
-
 // Print contents of the stream to cout in a thread-safe manner.
 // This class consumes stream inputs, buffers them, and writes them
 // out when destructed respecting the cout_mutex.
 class SafeCout {
   std::stringstream stream;
+  inline static std::mutex cout_mutex; // We need this to synchronise printing
 
 public:
   ~SafeCout() {
@@ -33,7 +32,6 @@ public:
     return *this;
   }
 };
-
 
 // A mock data object
 struct Data {
@@ -108,7 +106,7 @@ int main() {
 
     auto result = process(threadIdx, data);
 
-    SafeCout{} << '[' << threadIdx << "] Done " << (result ? "OK" : "with failure!") << '\n';
+    SafeCout{} << '[' << threadIdx << "] Data processing completed " << (result ? "OK" : "with failure!") << '\n';
   };
 
   std::vector<std::thread> consumers;
--- a/code/condition_variable/solution/condition_variable.sol.cpp
+++ b/code/condition_variable/solution/condition_variable.sol.cpp
@@ -13,13 +13,12 @@
  */
 using namespace std::chrono_literals; // We can write 1s
 
-std::mutex cout_mutex; // We need this to synchronise printing
-
 // Print contents of the stream to cout in a thread-safe manner.
 // This class consumes stream inputs, buffers them, and writes them
 // out when destructed respecting the cout_mutex.
 class SafeCout {
   std::stringstream stream;
+  inline static std::mutex cout_mutex; // We need this to synchronise printing
 
 public:
   ~SafeCout() {
@@ -33,7 +32,6 @@ public:
     return *this;
   }
 };
-
 
 // A mock data object
 struct Data {
@@ -108,7 +106,7 @@ int main() {
     }
     auto result = process(threadIdx, data);
 
-    SafeCout{} << '[' << threadIdx << "] Done " << (result ? "OK" : "with failure!") << '\n';
+    SafeCout{} << '[' << threadIdx << "] Data processing completed " << (result ? "OK" : "with failure!") << '\n';
   };
 
   std::vector<std::thread> consumers;
--- a/talk/concurrency/condition.tex
+++ b/talk/concurrency/condition.tex
@@ -2,9 +2,13 @@
 
 \begin{frame}[fragile]
   \frametitlecpp[11]{Condition variables}
-  \begin{block}{Communicating thread dependencies}
+  \begin{block}{Communicating between threads}
     \begin{itemize}
-      \item \mintinline{cpp}{std::condition_variable} from \mintinline{cpp}{<condition_variable>} header
+      \item Take the case where threads are waiting for other thread(s)
+      \item \mintinline{cpp}{std::condition_variable}
+      \begin{itemize}
+        \item from \mintinline{cpp}{<condition_variable>} header
+      \end{itemize}
       \item Allows for a thread to sleep (= conserve CPU time) until a given condition is satisfied
     \end{itemize}
   \end{block}
@@ -23,12 +27,14 @@
 \end{frame}
 
 \begin{frame}[fragile]
-  \frametitlecpp[17]{Using condition variables}
-  \begin{block}{Producer side}
+  \frametitlecpp[17]{Using condition variables: notify}
+  \begin{block}{Producer side: providing data to waiting threads}
     \begin{itemize}
-      \item Imagine multiple threads sharing data. Protect it with a mutex
-      \item Use a condition variable to notify consumers
-      \item Optimization: Don't hold lock while notifying (would block the waking threads)
+      \item Protect data with a mutex, and use condition variable to notify consumers
+      \item Optimal use: don't hold lock while notifying
+      \begin{itemize}
+        \item waiting threads would be blocked
+      \end{itemize}
     \end{itemize}
   \end{block}
   \begin{exampleblock}{}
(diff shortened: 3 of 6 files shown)


talk/concurrency/condition.tex
```diff
@@ -23,12 +24,11 @@
...
-  \frametitlecpp[17]{Using condition variables}
-  \begin{block}{Producer side}
+  \frametitlecpp[17]{Using condition variables: notify}
+  \begin{block}{Producer side: providing data to waiting threads}
     \begin{itemize}
-      \item Imagine multiple threads sharing data. Protect it with a mutex
-      \item Use a condition variable to notify consumers
-      \item Optimization: Don't hold lock while notifying (would block the waking threads)
+      \item Protect data with a mutex, and use condition variable to notify consumers
+      \item Optimal use: don't hold lock while notifying (waiting threads would be blocked)
```
sponce, line 31:
```suggestion
      \item Optimal use: don't hold lock while notifying
      \begin{itemize}
        \item waiting threads would be blocked
      \end{itemize}
```

talk/concurrency/condition.tex
```diff
@@ -2,8 +2,9 @@
 
 \begin{frame}[fragile]
   \frametitlecpp[11]{Condition variables}
-  \begin{block}{Communicating thread dependencies}
+  \begin{block}{Communicating between threads}
     \begin{itemize}
+      \item Imagine the case where threads are waiting for work from another thread
```
sponce, line 7:
```suggestion
      \item case where threads are waiting for (an)other thread(s)
```
That way it should fit on the line and look better

talk/concurrency/condition.tex
```diff
@@ -2,8 +2,9 @@
...
       \item \mintinline{cpp}{std::condition_variable} from \mintinline{cpp}{<condition_variable>} header
```
sponce, line 8:
```suggestion
      \item \mintinline{cpp}{std::condition_variable}
      \begin{itemize}
        \item from \mintinline{cpp}{<condition_variable>} header
      \end{itemize}
```

talk/concurrency/condition.tex
```diff
@@ -48,43 +48,51 @@
...
-      for (auto t : {&producer, &t1, &t2, ...}) t->join();
     \end{cppcode*}
+  \end{alertblock}
+\end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[11]{Using condition variables: correct wait}
+  \begin{block}{Waiting / waking up}
+    \begin{itemize}
+      \item {\color{green!80!black} Solution:} Put locking and waiting in a scope
```
sponce, line 90: Too light a color, try
```suggestion
      \item {\color{green!50!black} Solution:} Put locking and waiting in a scope
```

sponce (approved):
All fine ! I've only tried to fix some indentations here and there to make it look better. Also the green for "Solution" was really to light on my screen, especially on violet background

code/condition_variable/condition_variable.cpp
```diff
@@ -13,13 +13,12 @@
...
 using namespace std::chrono_literals; // We can write 1s
 
-std::mutex cout_mutex; // We need this to synchronise printing
-
 // Print contents of the stream to cout in a thread-safe manner.
 // This class consumes stream inputs, buffers them, and writes them
 // out when destructed respecting the cout_mutex.
 class SafeCout {
   std::stringstream stream;
+  static std::mutex cout_mutex; // We need this to synchronise printing
```
bernhardmgruber, line 21:
```suggestion
  inline static std::mutex cout_mutex; // We need this to synchronise printing
```

code/condition_variable/solution/condition_variable.sol.cpp
```diff
@@ -13,13 +13,12 @@
...
 using namespace std::chrono_literals; // We can write 1s
 
-std::mutex cout_mutex; // We need this to synchronise printing
-
 // Print contents of the stream to cout in a thread-safe manner.
 // This class consumes stream inputs, buffers them, and writes them
 // out when destructed respecting the cout_mutex.
 class SafeCout {
   std::stringstream stream;
+  static std::mutex cout_mutex; // We need this to synchronise printing
```
bernhardmgruber, line 21:
```suggestion
  inline static std::mutex cout_mutex; // We need this to synchronise printing
```

code/condition_variable/solution/condition_variable.sol.cpp
```diff
@@ -33,7 +32,8 @@ class SafeCout {
     return *this;
   }
 };
-
+// We need to explicitly initialise static members
+std::mutex SafeCout::cout_mutex{};
```
bernhardmgruber, line 36:
```suggestion
```

talk/concurrency/condition.tex
```diff
@@ -48,43 +48,51 @@
...
+  \begin{block}{Mechanics of wait}
     \begin{itemize}
-      \item Start many threads which have to wait for shared data
-      \item Provide a lock to be managed by \mintinline{cpp}{wait}
-      \item \mintinline{cpp}{wait} will only lock while necessary; unlocked while sleeping
-      \item Threads might wake up, but \mintinline{cpp}{wait} returns only when condition satisfied
+      \item Many threads are waiting for shared data
+      \item Pass a \mintinline{cpp}{unique_lock} and a predicate for wakeup to \mintinline{cpp}{wait()}
+      \item \mintinline{cpp}{wait()} sends threads to sleep while predicate is \mintinline{cpp}{false}
+      \item \mintinline{cpp}{wait()} will only lock while necessary; unlocked while sleeping
```
bernhardmgruber, line 60:
```suggestion
      \item \mintinline{cpp}{wait()} will only lock when necessary; unlocked while sleeping
```

talk/concurrency/condition.tex
```diff
@@ -48,43 +48,51 @@
...
+      \item Threads might wake up, but \mintinline{cpp}{wait()} returns only when lock available \emph{and} predicate \mintinline{cpp}{true}
```
bernhardmgruber, line 61: Should we mention the term `spurious wakeup` here?

code/condition_variable/condition_variable.cpp
hageboeck: Sure, I forgot the inline! 🤦‍♂️

talk/concurrency/condition.tex
hageboeck: Orally on the lecture? I can add it to the slides as well if you prefer.

talk/concurrency/condition.tex
hageboeck: I took it into account, but changed it a bit more. 🙂

talk/concurrency/condition.tex
bernhardmgruber: I meant mentioning on the slides in written form :)

talk/concurrency/mutexes.tex
```diff
@@ -184,10 +185,10 @@
       \begin{itemize}
       \item Or add master lock protecting the locking phase
       \end{itemize}
-    \item Respect a strict order in the locking across all threads
+    \item Respect a strict locking order across all threads
```
bernhardmgruber, line 188: You could mention that `std::scoped_lock` enforces uses a dead-lock avoidance algorithm if used on multiple mutexes.

talk/concurrency/threadsasync.tex
```diff
@@ -86,7 +86,7 @@
       \begin{itemize}
       \item execution starts when \mintinline{cpp}{get()} is called on the returned future
       \end{itemize}
-    \item default is not specified!
+    \item default is not specified, but tries to use existing concurrency!
```
bernhardmgruber, line 89: What does that mean? :) What is "existing concurrency"? Will it reuse the existing thread, which was spawned before somewhere else? Will it use an existing OpenMP threadpool? (probably not).

Could you maybe phrase this a bit more specifically. At least I would not know what to explain here :)

bernhardmgruber (approved):

talk/concurrency/threadsasync.tex
hageboeck: From [cppreference](https://en.cppreference.com/w/cpp/thread/async):
> If more than one flag is set, it is implementation-defined which policy is selected. For the default (both the [std::launch::async](https://en.cppreference.com/w/cpp/thread/launch) and [std::launch::deferred](https://en.cppreference.com/w/cpp/thread/launch) flags are set in policy), standard recommends (but doesn't require) utilizing available concurrency, and deferring any additional tasks

hageboeck: My takeaway is that it will use existing threads if possible, and defer other tasks. If TBB is at the backend, it will most likely just schedule a task and let TBB handle the rest.

talk/concurrency/mutexes.tex
hageboeck: You mean like 5 lines above?
> `\item \cpp17: \mintinline{cpp}{std::scoped_lock lock{m1, m2};} comes with deadlock-avoidance algorithm`

Yes, I agree! 😂

bernhardmgruber: Yeah!

PR #324, opened 2022-10-10
Tiny fixes after rehearsal of advanced course

sponce:
things like fixing line numbers, missing semi-colon.

Only 2 real changes : 
  - constinit slide
  - dynamic_cast example

### DIFF
--- a/talk/basicconcepts/classenum.tex
+++ b/talk/basicconcepts/classenum.tex
@@ -112,7 +112,7 @@
       VehicleType t = CAR;
     \end{cppcode*}
     \columnbreak
-    \begin{cppcode*}{gobble=2}
+    \begin{cppcode*}{gobble=2,firstnumber=8}
       enum VehicleType
         : int { // C++11
         BIKE = 3,
@@ -139,7 +139,7 @@
     \item strong typing, no automatic conversion to int
     \end{itemize}
     \small
-    \begin{cppcode*}{}
+    \begin{cppcode*}{firstnumber=3}
       enum VType { Bus, Car }; enum Color { Red, Blue };
       VType t = Bus;
       if (t == Red) { /* We do enter */ }
@@ -210,7 +210,7 @@
     \end{cppcode*}
   \end{alertblock}
   \begin{exampleblock}{\cpp11}
-    \begin{cppcode*}{gobble=2}
+    \begin{cppcode*}{gobble=2,firstnumber=4}
       using myint = uint64_t;
       myint toto = 17;
       using pos = int[3];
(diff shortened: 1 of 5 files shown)


talk/morelanguage/constexpr.tex
```diff
@@ -150,7 +150,7 @@
   \begin{block}{Motivation}
     \begin{itemize}
     \item Like a \mintinline{cpp}{constexpr} variable, a \mintinline{cpp}{constinit} variable guarantees compile-time initialization, but can be modified afterwards
-    \item Only allowed for \mintinline{cpp}{static}/global and \mintinline{cpp}{thread_local} variables
+    \item Makes variable \mintinline{cpp}{static}/global or \mintinline{cpp}{thread_local}
```
bernhardmgruber, line 153: This is not correct. `constinit` does not make a non-static variable `static`. E.g. This code:
```c++
int f() {
    constinit int a = 3; // error
}
struct S {
    constinit int a = 4; // error
};
```
Has two compile errors. So the original sentence was more correct. The `constinit` keyword is only allowed on a variable which is already `static`/global or `thread_local`.

bernhardmgruber (changes requested):

talk/morelanguage/constexpr.tex
sponce: hum... I've reread cppreference about that in the afternoon, and the very first sentence for constinit is : "The constinit specifier declares a variable with static or thread storage duration". See https://en.cppreference.com/w/cpp/language/constinit

And the example given does not use any static or anything.

talk/objectorientation/adl.tex
```diff
@@ -209,7 +209,7 @@
         \item Address-of \mintinline{cpp}{operator&}
         \item Boolean logic operators, e.g.\ \mintinline{cpp}{operator||}
         \item Comma operator \mintinline{cpp}{operator,}
-        \item Member dereference operators \mintinline{cpp}{operator->*}
+        \item Member dereference operators \mintinline{cpp}{operator->}
```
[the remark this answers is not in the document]
bernhardmgruber: But `operator.*` is not overloadable. That's why it is not mentioned here :)
These operators would be part of the member function pointers category. We could add a slide at some point: #326.

talk/morelanguage/constexpr.tex
bernhardmgruber: > And the example given does not use any static or anything.

The example on the slide is fine because the variables `a`, `b` and `c` are in global scope and thus have static storage duration.

sponce: ok, let me revert that too then

bernhardmgruber (approved):
I did not check the numbering of code snippets. The rest looks good!

PR #329, opened 2022-10-11
Fix delegating ctor example

eguiraud:


bernhardmgruber:
How does this fix the example? I think both versions (the one before and your suggested one) are valid use cases. I would rather prefer to either *add* your version, or add a comment that both situations are possible.

Delegating constructors are about avoiding code duplications. That should be the main point. How ctors delegate to each other is less important.

eguiraud:
Unfortunately the original example is not a valid use of delegating constructors (although I wish it was, because it makes absolute sense): https://godbolt.org/z/MTsz4n3x6

talk/objectorientation/constructors.tex
```diff
@@ -204,8 +204,8 @@
     \begin{cppcode}
       struct Delegate {
         int m_i;
-        Delegate() { ... complex initialization ...}
-        Delegate(int i) : Delegate(), m_i(i) {}
+        Delegate(int i) : m_i(i) {}
```
sponce, line 207:
```suggestion
        Delegate(int i) : m_i(i) {
          ... complex initialization ...
        }
```

sponce (approved):
Damn, I'm learning something... As every day with C++ ;-)

Anyway, I agree on the change but would still leave the "complex initialization" part just to insist on code reuse, which is the coere feature as @bernhardmgruber was pointing it

bernhardmgruber:
Wow, I did not know that! Cool!

PR #332, opened 2022-10-11
Improve wording about uninitialized values

eguiraud:
- Clarify that only ~~primitive~~ fundamental types can be uninitialized: the previous wording seemed to indicate that any object of a type T that does not have a default constructor can be left uninitialized.
- ~~Remove~~ Reword wrong statement about array initialization

### DIFF
--- a/talk/morelanguage/initialization.tex
+++ b/talk/morelanguage/initialization.tex
@@ -98,8 +98,8 @@
     \end{cppcode*}
     \begin{itemize}
       \item If \mintinline{cpp}{T} has a default constructor (including compiler generated), calls it
-      \item Otherwise \mintinline{cpp}{t} is left uninitialized (undefined value)
-      \item If \mintinline{cpp}{T} is an array, default initializes all members
+      \item If \mintinline{cpp}{T} is a fundamental type, \mintinline{cpp}{t} is left uninitialized (undefined value)
+      \item If \mintinline{cpp}{T} is an array, the same rules are applied to each item
     \end{itemize}
   \end{block}
 \end{frame}


talk/morelanguage/initialization.tex
```diff
@@ -98,8 +98,7 @@
     \end{cppcode*}
     \begin{itemize}
       \item If \mintinline{cpp}{T} has a default constructor (including compiler generated), calls it
-      \item Otherwise \mintinline{cpp}{t} is left uninitialized (undefined value)
-      \item If \mintinline{cpp}{T} is an array, default initializes all members
+      \item If \mintinline{cpp}{T} is a primitive type, \mintinline{cpp}{t} is left uninitialized (undefined value)
```
bernhardmgruber, line 101:
```suggestion
      \item If \mintinline{cpp}{T} is a fundamental type, \mintinline{cpp}{t} is left uninitialized (undefined value)
      \item If \mintinline{cpp}{T} is an array, default initializes all members
```
Please keep the point on arrays!

eguiraud: see the commit message, I think the about arrays is wrong? https://godbolt.org/z/6Paf5P49x

sponce: Testing it further in godbolt, it looks like the correct statement is probably "If \mintinline{cpp}{T} is an array, same rules are applied to each item"

eguiraud: fair enough :)

eguiraud: suggestions applied

bernhardmgruber (approved):

talk/morelanguage/initialization.tex
eguiraud: oooh yeah now I get it, sorry :/

PR #333, opened 2022-10-11
Items found on day 1 of advanced course

hageboeck:
This is a collections of things we brought up during the lecture.

### DIFF
--- a/talk/concurrency/threadlocal.tex
+++ b/talk/concurrency/threadlocal.tex
@@ -1,0 +1,26 @@
+\subsection[TLS]{Thread-local storage}
+
+\begin{frame}[fragile]
+  \frametitlecpp[20]{Thread-local storage}
+  \begin{block}{\mintinline{cpp}{thread_local} keyword}
+    \begin{itemize}
+      \item A variable can be declared \mintinline{cpp}{thread-local}
+      \item Then every thread will have its own copy
+      \item Most useful for ``working memory'' in each thread
+      \item Note: Can also be declared directly on the stack of a thread
+      \begin{itemize}
+        \item and will be faster, thus should be preferred when possible
+      \end{itemize}
+    \end{itemize}
+  \end{block}
+  \begin{exampleblock}{}
+    \begin{cppcode*}{}
+      thread_local int a{0};
+      std::jthread t1([&] { a++; });
+      std::jthread t2([&] { a++; });
+      a += 2;
+      t1.join(); t2.join();
+      assert( a == 2 ); // Guaranteed to succeed
+    \end{cppcode*}
+  \end{exampleblock}
+\end{frame}
--- a/talk/C++Course.tex
+++ b/talk/C++Course.tex
@@ -16,7 +16,7 @@
 \else
   % advanced slides have a different structure color
   \specialcomment{advanced}{
-    \setbeamercolor{structure}{fg=beamer@blendedblue!60!violet}
+    \setbeamercolor{structure}{fg=beamer@blendedblue!40!violet}
   }{
     \setbeamercolor{structure}{fg=beamer@blendedblue}
   }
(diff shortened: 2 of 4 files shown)


talk/concurrency/threadlocal.tex
```diff
@@ -0,0 +1,22 @@
+\subsection[TLS]{Thread-local storage}
+
+\begin{frame}[fragile]
+  \frametitlecpp[11]{Thread-local storage}
+  \begin{block}{\mintinline{cpp}{thread_local} keyword}
+    \begin{itemize}
+      \item When declaring a variable thread-local, every thread will have its own copy
+      \item Most useful for things treated as ``working memory'' in each thread
+      \item Note: Can also be declared directly on the stack of a thread (faster)
```
sponce, line 9:
```suggestion
      \item Note: Can also be declared directly on the stack of a thread
      \begin{itemize}
        \item and will be faster, thus should be preferred when possible
      \end{itemize}
```

talk/concurrency/threadlocal.tex
```diff
@@ -0,0 +1,22 @@
+\subsection[TLS]{Thread-local storage}
+
+\begin{frame}[fragile]
+  \frametitlecpp[11]{Thread-local storage}
+  \begin{block}{\mintinline{cpp}{thread_local} keyword}
+    \begin{itemize}
+      \item When declaring a variable thread-local, every thread will have its own copy
+      \item Most useful for things treated as ``working memory'' in each thread
```
sponce, line 8:
```suggestion
      \item Most useful for ``working memory'' in each thread
```

talk/concurrency/threadlocal.tex
```diff
@@ -0,0 +1,22 @@
+\subsection[TLS]{Thread-local storage}
+
+\begin{frame}[fragile]
+  \frametitlecpp[11]{Thread-local storage}
+  \begin{block}{\mintinline{cpp}{thread_local} keyword}
+    \begin{itemize}
+      \item When declaring a variable thread-local, every thread will have its own copy
```
sponce, line 7:
```suggestion
      \item A variable can be declared \mintinline{cpp}{thread-local}
      \item Then every thread will have its own copy
```

sponce (approved):
Just tried to make the thread-local slide nicer, but all fine aside from that

talk/concurrency/threadlocal.tex
```diff
@@ -0,0 +1,22 @@
...
+      \item Most useful for things treated as ``working memory'' in each thread
+      \item Note: Can also be declared directly on the stack of a thread (faster)
+    \end{itemize}
+  \end{block}
+  \begin{exampleblock}{}
+    \begin{cppcode*}{}
+      thread_local int a{0};
+      std::thread t1([&](){ a++; });
+      std::thread t2([&](){ a++; });
```
eguiraud, line 16: might also want to use a `jthread` instead 
```suggestion
      std::thread t1([&] { a++; });
      std::thread t2([&] { a++; });
```

talk/concurrency/threadlocal.tex
eguiraud: you might also want to mention that `thread_local` variables are (usually) implicitly `static` -- if you think about it it must be that way, but if you don't think about it you'll get surprising behavior :)

talk/concurrency/threadlocal.tex
hageboeck: You are generally right, but the example is supposed to remind students of other examples which we are showing. There, we don't use `jthread` yet.
On second thought, though, I'll accept the suggestion, since this gives us the opportunity to show it in action.

talk/concurrency/threadlocal.tex
eguiraud: `thread_local` implies `static` for variables at block scope (which in my experience is the situation in which you most commonly use `thread_local`)

PR #334, opened 2022-10-11
Minor fixes to perfect forwarding section

eguiraud:
- Remove a wrong statement: rvalues would actually be supported if wrapper took a constT&
- Preserve value type in wrapper function

### DIFF
--- a/talk/expert/perfectforwarding.tex
+++ b/talk/expert/perfectforwarding.tex
@@ -57,7 +57,6 @@
   \begin{alertblock}{}
     \begin{itemize}
       \item \mintinline{cpp}{const T&} won't work when passing something non const
-      \item rvalues are not supported in either case
     \end{itemize}
   \end{alertblock}
 \end{frame}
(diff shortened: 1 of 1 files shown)


sponce (approved):

talk/expert/perfectforwarding.tex
```diff
@@ -57,7 +57,6 @@
   \begin{alertblock}{}
     \begin{itemize}
       \item \mintinline{cpp}{const T&} won't work when passing something non const
-      \item rvalues are not supported in either case
```
bernhardmgruber, line 60: Why is this removed? It should be at least reworded in: `rvalues are also not supported`. In the example `wrapper(42)`, `42` is an rvalue (a prvalue), and cannot be bound to a mutable reference. This is important here, because otherwise `wrapper(T& arg)` could handle all cases.

bernhardmgruber: True. I would prefer to have that in the summary of the slide at the bottom though. But if you insist, we can leave it as is.

Issue #283, opened 2022-09-24
Add slides on function templates and template argument deduction


bernhardmgruber:
In the templates section we do not cover function templates. We only show a generic `max` as an example. This should be extended and especially include how template argument deduction works.

 In my variadic exercise, I have the following snipped:
```c++
template <std::size_t I, typename... Ts>
auto& get(tuple<Ts...>& t) {
    if constexpr (I == 0)
        return t.m_head;
    else
        return get<I - 1>(tail(t));
}
```
Especially the call to `get<I - 1>(tail(t))` may be difficult to understand, because one template argument needs to be provided and the other one is deduced.

PR #336, opened 2022-10-11
Extend slides on templates

bernhardmgruber:
I know this comes in super late, so feel free to exclude it for tomorrow.

Fixes: #283

### DIFF
--- a/talk/morelanguage/templates.tex
+++ b/talk/morelanguage/templates.tex
@@ -14,7 +14,7 @@
   \begin{cppcode}
     template<typename T>
     const T & max(const T &a, const T &b) {
-      return a > b ? a : b;
+      return b < a ? a : b;
     }
     template<typename T>
     struct Vector {
@@ -33,9 +33,10 @@
       \item they are compiled for each instantiation
       \item they need to be defined before used
       \begin{itemize}
-        \item so all templated code has to be in headers
+        \item so all template code must typically be in headers
+        \item or declared to be available externally (\mintinline{cpp}{extern template})
       \end{itemize}
-      \item this may lead to longer compilation times and bigger libraries
+      \item this may lead to longer compilation times and bigger binaries
     \end{itemize}
   \end{alertblock}
   \newsavebox{\codepiece}
@@ -104,7 +105,7 @@
 \end{frame}
 
 \begin{frame}[fragile]
-  \frametitlecpp[98]{Templates implementation}
+  \frametitlecpp[98]{Template implementation}
   \begin{cppcode*}{}
     template<typename KeyType=int, typename ValueType=KeyType>
     struct Map {
@@ -131,7 +132,7 @@
     \begin{itemize}
     \item integral types, pointer, enums in \cpp98
     \item \mintinline{cpp}{auto} in \cpp17
-    \item floats and literal types in \cpp20
+    \item floating-point and literal types in \cpp20
     \end{itemize}
   \end{block}
   \begin{cppcode*}{}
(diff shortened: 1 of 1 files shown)


talk/morelanguage/templates.tex
```diff
@@ -145,27 +146,213 @@
...
+\end{frame}
+
+\begin{advanced}
+
+\begin{frame}[fragile]
+  \frametitlecpp[98]{Template argument deduction}
+  \begin{block}{Template argument deduction}
+    \begin{itemize}
+    \item Only for function templates (Before \cpp14)
+    \item Template arguments deduced from function arguments
```
hageboeck, line 180: First say what it is, and then mention the restriction?

talk/morelanguage/templates.tex
```diff
@@ -145,27 +146,213 @@
...
+    \onslide<2>
+    \begin{alertblock}{}
+      \footnotesize
+      \begin{cppcode*}{gobble=2}
+        template <typename C, typename F>
+        void reduce(const C& cont, const F& f = std::plus{});
+        reduce(std::vector<int>{...}); // error: cannot deduce F
+        // would need: <typename C, typename F = decltype(std::plus{})>
+
+        const T & max(const T &a, const T &b) { ... }
```
hageboeck, line 254:
```suggestion
        template<typename T>
        const T & max(const T &a, const T &b) { ... }
```

talk/morelanguage/templates.tex
```diff
@@ -145,27 +146,213 @@
...
+
+    S<int>  s1; // prim. tmpl. (T=int) ok, spec. 1/2 invalid
+    S<int*> s2; // prim. tmpl. (T=int*) and spec. 1 (T=int) ok
+                // spec. 1 is more specialized, will be used
   \end{cppcode*}
 \end{frame}
 
+\begin{frame}[fragile]
+  \frametitlecpp[98]{Template argument deduction}
+  \begin{block}{Specialization vs. overloading}
```
hageboeck, line 294:
```suggestion
  \begin{block}{Specialization vs.\ overloading}
```

talk/morelanguage/templates.tex
```diff
@@ -145,27 +146,213 @@
...
+    \end{cppcode*}
+    \columnbreak
+    \begin{cppcode*}{}
+      template <typename T>
+      void f(T t) { ... }
+
+      template <>
+      void f<int*>(int* t) { ... }
+
+      // part. spec. forbidden:
```
hageboeck, line 320: Should we add why it's forbidden?
We could add this as a bullet point in the box above.

talk/morelanguage/templates.tex
```diff
@@ -145,27 +146,213 @@
...
+      template <typename T>
+      void f<T*>(T* t) {...}
+    \end{cppcode*}
+  \end{multicols}
+\end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[11]{Template argument deduction}
+  \begin{block}{Disadvantages of specialization vs. overloading}
```
hageboeck, line 329:
```suggestion
  \begin{block}{Disadvantages of specialization vs.\ overloading}
```

hageboeck (changes requested):
Here's a few suggestions. I believe I found a missing `template<...>`

talk/morelanguage/templates.tex
bernhardmgruber: So, what is the reason why it is forbidden? :)
The only reasons I know is that it would burden implementers and you can achieve similar effects with overloading.

bernhardmgruber: I just added a bullet point that it is forbidden.

talk/morelanguage/templates.tex
bernhardmgruber: yeah, totally forgot that!

talk/morelanguage/templates.tex
bernhardmgruber: ok!

PR #337, opened 2022-10-11
Minor improvements to modules and three-way comparison operator sections 

eguiraud:


talk/expert/cpp20spaceship.tex
```diff
@@ -39,7 +39,7 @@
       \item similar to \mintinline{cpp}{std::strcmp}
       \end{itemize}
     \item greater, lower or equal to \mintinline{cpp}{0} means respectively \emph{lower than}, \emph{greater than} and \emph{equivalent to}
-    \item It is provided by default for all predefined types
+    \item It is provided by default for all built-in types
```
bernhardmgruber, line 42: Consider adding:
```suggestion
    \item It is provided by default for all built-in types and many types in the standard library
```

talk/expert/modules.tex
```diff
@@ -401,7 +401,7 @@
       \begin{itemize}
         \item I.e.\ the header does not require setup of preprocessor state before inclusion
       \end{itemize}
-      \item All \cpp20 headers are importable
+      \item All standard library headers are importable
```
bernhardmgruber, line 404: Oh wait, this should still read `All \cpp20 standard library headers`. Because (as in the next sentence) the C standard library headers are not importable.

bernhardmgruber: I will add this to a subsequent PR.

eguiraud: you're right, didn't think about the C headers!

PR #356, opened 2022-10-12
Prefer unique_ptr<T[]> to manual memory management in templates ex

eguiraud:
It makes code "more RAII" and removes the need to implement a dtor.

I'd use `make_unique` instead of `new` but the Makefile explicitly builds with `-std=c++11` so I don't have `make_unique`. I don't know if we can bump it to C++14 instead.

I realize that this patch might be controversial for some, but I feel like an advanced C++ course should show that you basically never need to do manual memory management anymore in 2022.

### DIFF
--- a/code/templates/CMakeLists.txt
+++ b/code/templates/CMakeLists.txt
@@ -3,6 +3,9 @@
 cmake_minimum_required( VERSION 3.12 )
 project( templates LANGUAGES CXX )
 
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
 # Set up the compilation environment.
 include( "${CMAKE_CURRENT_SOURCE_DIR}/../CompilerSettings.cmake" )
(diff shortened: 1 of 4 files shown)


theanalyst (approved):

bernhardmgruber:
I am afraid if we merge this now and some students pull today/tomorrow, they will get conflicts in their working copies. So let's maybe wait to merge until tomorrow.

eguiraud:
The exercise now requires C++17 as per the comments above, so we are using `make_unique` instead of manual `new`s.

sponce:
Just for the historical background : the initial idea of the course was to only use in examples/exercises what was already presented. Thus no unique_ptr here as it was coming later in the course compared to templates.

Now in the meantime, we've extended, split and refactored the course. On top, I fully share the argument that we should not show old/bad/pointer code and that it's better telling people "just don't worry about unique_ptr, it's just a better pointer and details come tomorrow".

So merging :-)

sponce (approved):

PR #357, opened 2022-10-12
Extend variants exercise

bernhardmgruber:
Added a function that counts the number of neutrons. This suggests the use of inspecting a variants currently held alternative.

### DIFF
--- a/code/variant/variant.cpp
+++ b/code/variant/variant.cpp
@@ -1,15 +1,17 @@
 /*
 
-In the code below, replace inheritance with the use of a std::variant.
-Two solutions are provided :
-1. with `std::get_if`,
-2. with `std::visit`.
+In the code below, replace inheritance and dynamic allocation (make_unique) with the use of a std::variant.
+Multiple solutions are possible. We provide solutions using:
+1. `std::get_if`,
+2. `std::visit`/`std::holds_alternative`.
 
 */
 
 #include <iostream>
 #include <memory>
 #include <vector>
+#include <algorithm>
+#include <span>
 
 struct Particle {
   virtual void print() const = 0;
@@ -28,6 +30,12 @@ struct Neutron : Particle {
   void print() const override { std::cout << "N\n"; }
 };
 
+std::size_t countNeutrons(std::span<const std::unique_ptr<Particle>> ps) {
+    return std::count_if(ps.begin(), ps.end(), [](const std::unique_ptr<Particle>& p){
+        return typeid(*p) == typeid(Neutron);
+    });
+}
+
 int main() {
   std::vector<std::unique_ptr<Particle>> ps;
   ps.push_back(std::make_unique<Electron>());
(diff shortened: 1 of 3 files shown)


code/variant/variant.cpp
```diff
@@ -28,6 +30,12 @@ struct Neutron : public Particle {
   virtual void print() const { std::cout << "N" << std::endl; }
 };
 
+std::size_t countNeutrons(std::span<const std::unique_ptr<Particle>> ps) {
+    return std::count_if(ps.begin(), ps.end(), [](const std::unique_ptr<Particle>& p){
+        return typeid(*p) == typeid(Neutron);
```
theanalyst, line 35: Would dynamic_cast be better here over RTTI even for showing problematic code, just to prevent someone from actually abusing this IRL ;)

bernhardmgruber: Hmm, I would implement it with `typeid`, which is less of a hammer than `dynamic_cast`. But AFAIK we also do not discuss `typeid` in the lecture, so maybe `dynamic_cast` is the better choice than. And sure, it is scary!

hageboeck (approved):
I'm happy with typeid. However, we might need a lot more instructions for a future version.
Would you agree? Issue?

sponce:
I'm not sure this addition brings much here. With typeid yes, but probably we should have dedicated slide and example. With get_if, it does not add much to the exercise providing we already add the first (non perfect) solution using it.

bernhardmgruber:
> I'm not sure this addition brings much here. With typeid yes, but probably we should have dedicated slide and example. With get_if, it does not add much to the exercise providing we already add the first (non perfect) solution using it.

You are probably right. I wanted to construct a case where `std::visit` is not the right tool and you have to use a different feature of `std::variant` to find out which type is currently held.

I leave it up to you! We can include this, or a similar function. Or we can just leave the exercise as is.

code/variant/variant.cpp
chavid: If the intent is to intriduce a better use-case for `get_if`, I think `dynamic_cast` is more suited, because having a similar mechanism, which is to return a null pointer if the expected type is not the real one. Also, the spirit of the exercice being the move from an "OO paradigm" to a "Functional paradigm", I think `dynamic_cast` is a lot more "typical OO idiom". Yes, it is very subjective ;)

theanalyst: Since we are moving down an inheritance hierarchy rather than 2 disjoint classes, I kind of feel like `dynamic_cast` is a better alternative like @chavid  mentions. Also the rare times you've had to inspect types at runtime a common case is when you'd have to invoke some specific method from the derived class in which case you're bound to `dynamic_cast` (which is ofc a bad design... but life;) 
Not a physicist, but for this example you could add something like a `get_antiparticle` which returns a class and since one can't do this anymore with inheritance you'd have the some derived classes implementing this or not.

This would also be a good use case for using variant of particles as it shows the use case for the times you'd want to unify types that don't need a hierarchy ;)

sponce:
Ok, I'm closing this without merging on the base of my last comment. Feel free to reopen if you think it's not the right move !

PR #359, opened 2022-10-12
Improve templates exercise

bernhardmgruber:
* Reword/extend README.md
* Repeat some instructions inside the code
* Use std::less instead of custom less
* Already provide the implementation of RevOrderString, so students can focus on the template related stuff
* Move to C++20


code/templates/playwithsort.cpp
```diff
@@ -31,4 +31,15 @@ int main() {
...
 
+    // TODO: Demonstrate OrderedVector with Complex as done above
+
+
+    // TODO: Extend OrderedVector to allow to customize the ordering via an additional template paramter.
+    //       Then, demonstrate the new functionality by ordering an OrderedVector<std::string>,
+    //       where the strings are compared starting at their last letters.
+
+
+    // TODO: Order an OrderedVector of Complex base don the Manhattan distance
```
hageboeck, line 42:
```suggestion
    // TODO: Order an OrderedVector of Complex based on the Manhattan distance
```

code/templates/playwithsort.cpp
```diff
@@ -3,13 +3,13 @@
...
 #include <iostream>
 #include <algorithm>
+#include <ranges>
 
-/*
-struct OrderFunctor {
-    bool operator() (const TYPE &s, const TYPE &t) const {
+struct RevOrderString {
+    bool operator() (const std::string &s, const std::string &t) const {
+        return std::ranges::lexicographical_compare(std::views::reverse(s), std::views::reverse(t));
```
hageboeck, line 10: Is this supposed to be here or only in the solution? If it's only supposed to be in the solution, I would nevertheless include some hints like "check the function std::ranges::lexicographical_compare for comparing strings.

hageboeck (approved):
Good after the two suggestions below.

hageboeck:
> Good after the two suggestions below.

Well, maybe not. Mac doesn't seem to speak "view" ...

code/templates/playwithsort.cpp
sponce: I would indeed put only a hint, so that people have to dig a bit cppreference. Having it fully empty as it was was probably a bit too hard though

sponce (approved):
Ok for me with an hint for the reverse order implementation

code/templates/playwithsort.cpp
bernhardmgruber: I thought I add the full implementation so people have

1. an example for how the comparison functor should look like
2. don't need to spend time to figure out how to implement the reverse string comparison and focus on the template stuff.
3. 
But it's not necessary, I can omit it.

sponce: You have a point that the goal of the exercise is not to dig cppreference and that it's already hard enough for students without implementing that. So ok, let's keep the implementation

bernhardmgruber:
I added a hint on how to implement the comparison functor instead of the implementation. I also renamed the comparison types to include `Less`, so it expresses that it implements a `less than` comparison.

bernhardmgruber:
Failure on MacOS:
```
/Users/runner/work/cpluspluscourse/cpluspluscourse/code/templates/solution/playwithsort.sol.cpp:11:58: error: no member named 'views' in namespace 'std'
```
Could be that the libc++ does not have a full ranges implementation there :/ I will fix that later.

PR #363, opened 2022-10-12
Improve SFINAE slides

bernhardmgruber:


talk/expert/sfinae.tex
```diff
@@ -6,37 +6,38 @@
...
   \end{exampleblock}
-  Note : SFINAE is largely superseded by concepts in \cpp20
+  \begin{alertblock}{}
+    Note: SFINAE is largely superseded by concepts in \cpp20
+  \end{alertblock}
 \end{frame}
 
 \begin{frame}[fragile]
-  \frametitlecpp[11]{decltype}
+  \frametitlecpp[11]{\texttt{decltype}}
```
sponce, line 27: I realize that the usage of decltype with lambdas yesterday could not be understood, as decltype was not known...
To be skipped in the advanced course, and reserved for an expert one, if any

talk/expert/sfinae.tex
```diff
@@ -218,55 +249,54 @@
...
 
 \begin{frame}[fragile]
-  \frametitlecpp[11]{Back to variadic templated class}
-  \begin{block}{The tuple get method}
+  \frametitlecpp[11]{Back to variadic class templates}
+  \begin{block}{The \texttt{tuple\_element} trait}
     \begin{cppcode*}{}
       template <size_t I, typename Tuple>
-      struct elem_type;
+      struct tuple_element;
```
sponce, line 262: I do not care so much about the name, but just as a historical note : originally this was really a cut and paste from the standard library

sponce (approved):

talk/expert/sfinae.tex
bernhardmgruber: Really? The type function is called [`tuple_element`](https://en.cppreference.com/w/cpp/utility/tuple/tuple_element) nowadays. I did not know it had a different name back then! :D

sponce: I confirm the latest code of the gcc implementation uses tuple_element

PR #366, opened 2022-10-13
Things found on day 3.

hageboeck:
- Add links to cppinsights for variadic templates


sponce:
Sorry but I'll close that one as part of it is handled better by #367 and the other part (f->g) will come in a moment with another bunch of tiny fixes discovered today

PR #367, opened 2022-10-13
Introduced easy linking to godbolt, cppinsight and cppreference in titles and started using them

sponce:


talk/expert/variadictemplate.tex
```diff
@@ -121,7 +121,7 @@
 
 \begin{frame}[fragile]
   \frametitlecpp[11]{Variadic class template}
-  \begin{block}{The tuple example, simplified}
+  \begin{block}{The tuple example, simplified \godboltLink{https://godbolt.org/z/nxjja4bob}}
```
hageboeck, line 124: I don't see how to go to cppinsights from godbolt. I had this is #366:
https://cppinsights.io/s/600802a9

sponce: Just click on the magnifier icon :-)

PR #368, opened 2022-10-13
More links to godbolt, cppinsight and cppreference

sponce:
Depends on #367

### DIFF
--- a/talk/concurrency/mutexes.tex
+++ b/talk/concurrency/mutexes.tex
@@ -2,7 +2,7 @@
 
 \begin{frame}[fragile]
   \frametitlecpp[11]{Races}
-  \begin{exampleblock}{Example code}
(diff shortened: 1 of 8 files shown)


sponce:
Does someone understand what is failing with this one. It's fine locally, and the output is so huge here (we seem to have issues with the tools used by the way) that I cannot find the error...

hageboeck (approved):
LGTM, but I didn't have the time to check all links.

hageboeck:
> Does someone understand what is failing with this one. It's fine locally, and the output is so huge here (we seem to have issues with the tools used by the way) that I cannot find the error...

I suspect a problem with the pdf conversion. There were more problems in other PRs, so it's probably unrelated. I'll try pdflatex or similar, and see if it's going any better.

PR #371, opened 2022-10-13
Use more inline definitions in templates exercise

bernhardmgruber:
This should reduce the amount of code a bit. Students also need to change the template heads in less places when adding the ordering predicate.


sponce (approved):
Yes, that's much better. Now let's wait a couple of days before emerging in case students are finishing the exercise

PR #373, opened 2022-10-14
Try pdflatex instead of xelatex

hageboeck:


sponce:
As a side note : the original issue was a use of minted in a title. I've replaced with texttt and it was fine. I still do not understand why it has been working locally yesterday (was failing today, that's how I found it).
Still as we use pdflatex locally (at least the Makefile does), it makes sense to also use it in the CI !

sponce (approved):

PR #375, opened 2022-10-14
Restrict workflows to subfolders.

hageboeck:
Only run code checks when code/ is touched; latex when talk/ is touched.


sponce:
Very good idea !

hageboeck:
> Very good idea !

I know, but now nothing is running. 😂

PR #377, opened 2022-10-14
Introduced exampleblockGB environment for examples with godbolt links

sponce:
And added also an index of godbolt code examples at the end with links both to the slide concerned and to godbolt

### DIFF
--- a/talk/setup.tex
+++ b/talk/setup.tex
@@ -106,7 +106,7 @@
 \makeatletter
 \newcommand\listofgoodpractices{%
   \addtocontents{lgp}{\end{enumerate}}
-  \small
+  \scriptsize
   \begin{multicols}{2}
     \@starttoc{lgp}%
   \end{multicols}
@@ -176,8 +176,6 @@
   \end{exerciseWithShortcut}
 }
 
-\makeatother
-
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 % frametitle with C++ version %
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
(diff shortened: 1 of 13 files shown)


talk/setup.tex
```diff
@@ -189,10 +187,38 @@
...
-  \href{#1}{\color{blue!50!white} godbolt}%
+\newcommand\godboltLink[2]{%
+  \ifnum \overlaynumber=1%
+    \href{#1}{\color{blue!50!white} godbolt}%
+    \hypertarget<1>{godbolt.\thegb@counter}{}
+    \addtocontents{lgb}{\protect\item \protect\hyperlink{godbolt.\thegb@counter}{#2 - \href{#1}{godbolt}\protect\hfill\insertframenumber}}%
+    \stepcounter{gb@counter}%
+  \fi
 }
+
```
hageboeck, line 213:
```suggestion

% Use: \begin{exampleblockGB}{<Box Title>}{<godbolt link>}{<title in index>}}
```

hageboeck (approved):
👍 
Maybe just add a comment on how to use it as suggested below?

talk/setup.tex
```diff
@@ -189,10 +187,38 @@
...
+\regtotcounter{gb@counter}
+\resetcounteronoverlays{gb@counter}
+\newcommand*{\overlaynumber}{\number\beamer@slideinframe}
+
 % Use as \godbolLink{url}
-\newcommand\godboltLink[1]{%
-  \href{#1}{\color{blue!50!white} godbolt}%
+\newcommand\godboltLink[2]{%
+  \ifnum \overlaynumber=1%
+    \href{#1}{\color{blue!50!white} godbolt}%
```
hageboeck, line 207:
```suggestion
  \href{#1}{\color{blue!50!white} godbolt}%
    \ifnum \overlaynumber=1%
```
Like this, the link will also show on the other slides, but the entry to the index only happens once.
However, if the box only appears on overlay != 1, there won't be an entry at all. Is this an issue?

sponce: Oh indeed ! Thanks for spotting.

hageboeck:
This seems to work now. I might copy the technique of inserting only on overlay number 1 for the list of good practices ...

PR #380, opened 2022-10-26
Update Makefile to use latexmk, and some minor improvements to the basic concepts section

amadio:
I added a `make preview` target to the `Makefile`, which should make it easier to avoid edit compile cycles with an appropriate viewer. On Linux, to use evince as your previewer, add the line below to `~/.latexmkrc`:
```
$pdf_previewer = "start evince"
```
I left it out of the custom config because it's likely different people will want to use different viewers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,17 +1,6 @@
 # latex related, Course
-talk/*aux
-talk/*/*aux
-C++Course.log
-C++Course.nav
-C++Course.out
-C++Course.snm
-C++Course.toc
-C++Course.vrb
-C++Course.xdv
-C++Course.pdf
-C++Course*.vrb
-C++Course.fdb_latexmk
-C++Course.fls
+C++Course.*
+!C++Course.tex
 _minted*
 
 # latex related, exercise intro
--- a/talk/basicconcepts/assert.tex
+++ b/talk/basicconcepts/assert.tex
@@ -77,7 +77,7 @@ Abort trap: 6
     \begin{cppcode*}{}
       double f(UserType a) {
         static_assert(
-          std::is_floating_point_v<UserType>,
+          std::is_floating_point<UserType>::value,
           "This function expects floating-point types.");
         return std::sqrt(a);
       }
@@ -89,7 +89,7 @@ a.cpp: In function 'double f(UserType)':
 \textcolor{blue}{a.cpp:3:9:} \textcolor{red}{error:} \textcolor{blue}{static assertion failed: This function}
                                            \textcolor{blue}{expects floating-point types.}
    2 | static_assert(
-     |   \textcolor{red}{std::is_floating_point_v<UserType>},
-     |   \textcolor{red}{~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~}
+     |   \textcolor{red}{std::is_floating_point<UserType>::value},
+     |   \textcolor{red}{~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~}
   \end{Verbatim}
 \end{frame}
--- a/talk/basicconcepts/assert.tex
+++ b/talk/basicconcepts/assert.tex
@@ -64,12 +64,12 @@ Abort trap: 6
 \end{frame}
 
 \begin{frame}[fragile]
-  \frametitlecpp[17]{Static Assert}
+  \frametitlecpp[11]{Static Assert}
   \begin{block}{Checking invariants at compile time}
     \begin{itemize}
       \item To check invariants at compile time, use \mintinline{cpp}{static_assert}
       \item The assertion can be any constant expression (see later)
-      \item No runtime penalty!
+      \item The message argument is optional in \cpp 17 and later
     \end{itemize}
   \end{block}
   \begin{exampleblock}{\texttt{static\_assert}}
@@ -86,7 +86,8 @@ Abort trap: 6
   \scriptsize
   \begin{Verbatim}[commandchars=\\\{\}]
 a.cpp: In function 'double f(UserType)':
-\textcolor{blue}{a.cpp:3:9:} \textcolor{red}{error:} \textcolor{blue}{static assertion failed: This function expects floating-point types.}
+\textcolor{blue}{a.cpp:3:9:} \textcolor{red}{error:} \textcolor{blue}{static assertion failed: This function}
+                                           \textcolor{blue}{expects floating-point types.}
    2 | static_assert(
      |   \textcolor{red}{std::is_floating_point_v<UserType>},
      |   \textcolor{red}{~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~}
--- a/talk/basicconcepts/auto.tex
+++ b/talk/basicconcepts/auto.tex
@@ -4,8 +4,8 @@
   \frametitlecpp[11]{Auto keyword}
   \begin{block}{Reason of being}
     \begin{itemize}
-    \item many type declarations are redundant
-    \item and lead to compiler errors if you mess up
+    \item Many type declarations are redundant
+    \item They are also a source for compiler warnings and errors
     \end{itemize}
     \begin{cppcode*}{}
       std::vector<int> v;
@@ -30,7 +30,7 @@
   \begin{exerciseWithShortcut}{Loops, references, auto}{Loops, refs, auto}
     Familiarise yourself with range-based for loops and references
     \begin{itemize}
-      \item go to \texttt{code/loopsRefsAuto}
+      \item Go to \texttt{code/loopsRefsAuto}
       \item Look at \texttt{loopsRefsAuto.cpp}
       \item Compile it (\texttt{make}) and run the program (\texttt{./loopsRefsAuto})
       \item Work on the tasks that you find in \texttt{loopsRefsAuto.cpp}
--- a/talk/basicconcepts/auto.tex
+++ b/talk/basicconcepts/auto.tex
@@ -5,7 +5,8 @@
   \begin{block}{Reason of being}
     \begin{itemize}
     \item Many type declarations are redundant
-    \item They are also a source for compiler warnings and errors
+    \item They are often a source for compiler warnings and errors
+    \item Using auto prevents unwanted/unnecessary type conversions
     \end{itemize}
     \begin{cppcode*}{}
       std::vector<int> v;
--- a/talk/basicconcepts/control.tex
+++ b/talk/basicconcepts/control.tex
@@ -62,10 +62,11 @@
     \end{cppcode*}
   \end{exampleblock}
   \pause
-  \begin{alertblock}{Do not abuse}
+  \begin{alertblock}{Do not abuse it}
     \begin{itemize}
-      \item explicit ifs are easier to read \\
-      \item use only when obvious and not nested
+      \item Explicit \mintinline{cpp}{if}s are generally easier to read
+      \item Use the ternary operator with short conditions and expressions
+      \item Avoid nesting
     \end{itemize}
   \end{alertblock}
 \end{frame}
--- a/talk/basicconcepts/control.tex
+++ b/talk/basicconcepts/control.tex
@@ -91,7 +91,7 @@
   \end{block}
   \pause
   \begin{alertblock}{Use break}
-    Do not try to make use of non breaking cases
+    Avoid \mintinline{cpp}{switch} statements with fall-through cases
   \end{alertblock}
 \end{frame}
 
--- a/talk/basicconcepts/control.tex
+++ b/talk/basicconcepts/control.tex
@@ -14,9 +14,9 @@
       }
     \end{cppcode*}
     \begin{itemize}
-      \item \mintinline{cpp}{else} and \mintinline{cpp}{else if} clause are optional
-      \item \mintinline{cpp}{else if} clause can be repeated
-      \item braces are optional if there is a single statement
+      \item The \mintinline{cpp}{else} and \mintinline{cpp}{else if} clauses are optional
+      \item The \mintinline{cpp}{else if} clause can be repeated
+      \item Braces are optional if there is a single statement
     \end{itemize}
   \end{block}
 \end{frame}
@@ -49,8 +49,8 @@
     \end{cppcode*}
     \vspace{-0.2cm}
     \begin{itemize}
-      \item if test is \mintinline{cpp}{true} expression1 is returned
-      \item else expression2 is returned
+      \item If test is \mintinline{cpp}{true} expression1 is returned
+      \item Else, expression2 is returned
     \end{itemize}
   \end{block}
   \pause
--- a/talk/basicconcepts/control.tex
+++ b/talk/basicconcepts/control.tex
@@ -83,10 +83,10 @@
       }
     \end{cppcode*}
     \begin{itemize}
-      \item \mintinline{cpp}{break} is not mandatory but...
-      \item cases are entry points, not independent pieces
-      \item execution falls through to the next case without a \mintinline{cpp}{break}!
-      \item \mintinline{cpp}{default} may be omitted
+      \item The \mintinline{cpp}{break} statement is not mandatory but...
+      \item Cases are entry points, not independent pieces
+      \item Execution falls through to the next case without a \mintinline{cpp}{break}!
+      \item The \mintinline{cpp}{default} case may be omitted
     \end{itemize}
   \end{block}
   \pause
@@ -142,7 +142,7 @@
 \end{frame}
 
 \begin{frame}[fragile]
-  \frametitlecpp[17]{init-statements for if and switch}
+  \frametitlecpp[17]{Init-statements for if and switch}
   \begin{block}{}
     Allows to limit variable scope in \mintinline{cpp}{if} and \mintinline{cpp}{switch} statements
   \end{block}
@@ -176,9 +176,9 @@
     \end{cppcode*}
     \vspace{-0.2cm}
     \begin{itemize}
-      \item initializations and increments are comma separated
-      \item initializations can contain declarations
-      \item braces are optional if loop body is a single statement
+      \item Initializations and increments are comma separated
+      \item Initializations can contain declarations
+      \item Braces are optional if loop body is a single statement
     \end{itemize}
   \end{block}
   \pause
@@ -201,8 +201,8 @@
   \frametitlecpp[11]{Range-based loops}
   \begin{block}{Reason of being}
     \begin{itemize}
-    \item simplifies loops over ``ranges'' tremendously
-    \item especially with STL containers
+    \item Simplifies loops over ``ranges'' tremendously
+    \item Especially with STL containers
     \end{itemize}
   \end{block}
   \begin{block}{Syntax}
@@ -222,7 +222,7 @@
 \end{frame}
 
 \begin{frame}[fragile]
-  \frametitlecpp[20]{init-statements for range-based loops}
+  \frametitlecpp[20]{Init-statements for range-based loops}
   \begin{block}{}
     Allows to limit variable scope in range-based loops
   \end{block}
@@ -257,7 +257,7 @@
       } while(condition);
     \end{cppcode*}
     \begin{itemize}
-      \item braces are optional if the body is a single statement
+      \item Braces are optional if the body is a single statement
     \end{itemize}
   \end{block}
   \pause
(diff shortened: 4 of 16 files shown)


talk/basicconcepts/control.tex
```diff
@@ -75,13 +75,13 @@
...
-        ...
-        default : instructiond; break;
-      }
+        switch(identifier) {
+          case c1 : statements1; break;
+          case c2 : statements2; break;
+          case c3 : statements3; break;
+          ...
+          default : instructiond; break;
+        }
```
hageboeck, line 84: I think you could do the same using `gobble=0`. To be tried.

talk/basicconcepts/assert.tex
```diff
@@ -86,9 +86,10 @@
   \scriptsize
   \begin{Verbatim}[commandchars=\\\{\}]
 a.cpp: In function 'double f(UserType)':
-\textcolor{blue}{a.cpp:3:9:} \textcolor{red}{error:} \textcolor{blue}{static assertion failed: This function expects floating-point types.}
+\textcolor{blue}{a.cpp:3:9:} \textcolor{red}{error:} \textcolor{blue}{static assertion failed: This function}
+                                           \textcolor{blue}{expects floating-point types.}
    2 | static_assert(
      |   \textcolor{red}{std::is_floating_point_v<UserType>},
-     |   \textcolor{red}{~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~}
+     |   \textcolor{red}{~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~}
```
hageboeck, line 93: My compiler really points to the `i` using a `^`. If you remove it, you can remove the line with the `~~~` as well.

hageboeck (approved):
LGTM.

I added two small comments below.

talk/basicconcepts/assert.tex
amadio: I will add it back, I thought this might have been a typo.

talk/basicconcepts/control.tex
amadio: I will rebase and use `gobble=0` instead.

.gitignore
```diff
@@ -1,17 +1,6 @@
...
-C++Course.snm
-C++Course.toc
-C++Course.vrb
-C++Course.xdv
-C++Course.pdf
-C++Course*.vrb
-C++Course.fdb_latexmk
-C++Course.fls
+C++Course.*
+!C++Course.tex
```
bernhardmgruber, line 3: Wow, I learned something!

talk/basicconcepts/assert.tex
```diff
@@ -64,12 +64,12 @@
 \end{frame}
 
 \begin{frame}[fragile]
-  \frametitlecpp[17]{Static Assert}
+  \frametitlecpp[11]{Static Assert}
```
bernhardmgruber, line 67: I we had the 17 because of `std::is_floating_point_v`. I know it's confusing because we discuss `static_assert` :S

talk/basicconcepts/control.tex
```diff
@@ -49,23 +49,24 @@
     \end{cppcode*}
     \vspace{-0.2cm}
     \begin{itemize}
-      \item if test is \mintinline{cpp}{true} expression1 is returned
-      \item else expression2 is returned
+      \item If test is \mintinline{cpp}{true} expression1 is returned
+      \item Else expression2 is returned
```
bernhardmgruber, line 53: Since you touched it. Maybe:
```suggestion
      \item Else, expression2 is returned
```

talk/basicconcepts/control.tex
```diff
@@ -49,23 +49,24 @@
...
   \end{block}
   \pause
   \begin{exampleblock}{Practical example}
     \begin{cppcode*}{}
-      int collatz(int a) {
-        return a==1 ? 1 : collatz(a%2==0 ? a/2 : 3*a+1);
-      }
+     unsigned int collatz(unsigned int a) {
+      return a == 1 ? 1 : collatz(a & 1 ? 3*a + 1 : a/2);
+     }
```
bernhardmgruber, line 61: I am not sure this version is better. Compilers will generate the bit op for you from the modulo op. I would argue the modulo is easier to read as. I guess the `unsigned` was chosen because of the bit op.

talk/basicconcepts/auto.tex
```diff
@@ -4,8 +4,8 @@
   \frametitlecpp[11]{Auto keyword}
   \begin{block}{Reason of being}
     \begin{itemize}
-    \item many type declarations are redundant
-    \item and lead to compiler errors if you mess up
+    \item Many type declarations are redundant
+    \item They are also a source for compiler warnings and errors
```
bernhardmgruber, line 8: You may want to add: unnecessary conversions.

Especially alternating between float/double or int, unsigned, size_t, because that's what you thought the type was, can sometimes create actual conversion code the CPU needs to perform. Or may choose the wrong overload of `std::sin`.

talk/basicconcepts/functions.tex
```diff
@@ -181,11 +181,11 @@
   \frametitlecpp[98]{Pass by value, reference or pointer}
   \begin{block}{Different ways to pass arguments to a function}
     \begin{itemize}
-    \item by default, arguments are passed by value (= copy) \\
-          good for small types, e.g.\ numbers
-    \item prefer references for mandatory parameters to avoid copies
-    \item use pointers for optional parameters to allow \mintinline{cpp}{nullptr}
-    \item use \mintinline{cpp}{const} for safety and readability whenever possible
+    \item By default, arguments are passed by value (= copy) \\
+          good for small types, e.g.\ anything smaller than a pointer
```
bernhardmgruber, line 185: Make it a bit larger please :) I think we argue later to also pass `std::span` and `std::string_view` by value, which are a pointer and a size_t. So I guess the guideline is: Pass by value what fits in a few registers.

I am also frequently passing `vecgeom::Vector<double>` by value now or `std::complex`. I think it follows a similar reasoning. But you can debate me! :)

talk/basicconcepts/references.tex
```diff
@@ -29,24 +29,23 @@
   \frametitlecpp[98]{Pointers vs References}
   \begin{block}{Specificities of reference}
     \begin{itemize}
-    \item natural syntax
-    \item must be assigned when defined, cannot be \mintinline{cpp}{nullptr}
-    \item cannot be reassigned
-    \item non-const references to temporary objects are not allowed
+    \item Natural syntax
+    \item Cannot be \mintinline{cpp}{nullptr}
+    \item Must be assigned when defined, cannot be reassigned
```
bernhardmgruber, line 34: When talking about references, there is also to term "to reseat" a reference/pointer instead of reassign or reset. Should we mention it somewhere is it too much C++ lingo?

talk/basicconcepts/references.tex
```diff
@@ -29,24 +29,23 @@
...
     \end{itemize}
   \end{block}
   \pause
   \begin{goodpractice}{References}
     \begin{itemize}
-      \item Always use references when you can
-      \item Consider that a referenced object can be modified
-      \item Use \mintinline{cpp}{const} when it's not the case
+      \item Prefer using references instead of pointers
+      \item Mark references \mintinline{cpp}{const} to prevent modification
```
bernhardmgruber, line 48: Since you removed the "Consider that a referenced object can be modified": Should we still add something along the lines of: "Remember that non-const references/pointers can modify the referred object" (e.g. in the context of a function you pass that reference/pointer to) or is that just too obvious?

talk/basicconcepts/assert.tex
amadio: Ah, indeed! I remember discussing that with you now. Still, the slide is about `static_assert`, not `std::is_floating_point_v`, so using C++17 gives the impression that `static_assert` itself is C++17 only. I think I will replace `_v` with `::value` then, what do you think?

talk/basicconcepts/auto.tex
amadio: Yes, I will add this to the slides if it fits.

talk/basicconcepts/control.tex
amadio: Most compilers optimize both versions to `return 1`, funnily enough. However, when they don't, my version has slightly shorter assembly. https://godbolt.org/z/5c46cnYYf In any case, I just replaced `a%2==0` with `a & 1` so that I could put spaces to format the code a bit better. Since bitwise operators are discussed nearby in the contents, I thought this would be an ok tradeoff to make for readability. The unsigned was chosen because a check for `a < 0` was removed compared with the previous slide. Using unsigned makes that impossible (this is in the commit message).

talk/basicconcepts/functions.tex
amadio: Yes, I tend to agree with you, but I wanted to give in a few words something beginners would easily understand, while having the [ABI](https://refspecs.linuxbase.org/elf/x86_64-abi-0.99.pdf#page=18) in mind. We can specify a size (i.e. up to 16 or 32 bytes), but I think that the size of a pointer is useful as advice because it's the size of a reference as well, and depending on what you are passing, it will be passed by reference anyway even when passed by value in code. I think I will also try to add a link to the ABI to the slide, for those who may want to check the details.

talk/basicconcepts/references.tex
amadio: I think it's out of scope for this pull request, but in any case I think that for the basic part, it's better to leave that out.

talk/basicconcepts/references.tex
amadio: Well, the last item tells people to mark as `const` what they don't want to be modified, which implies that if you don't mark them as `const`, they *can* be modified. Maybe this is not too obvious to beginners. I can add back that point if you want, I removed because the slide was too full with 3 items and I thought it was a bit redundant.

talk/basicconcepts/assert.tex
bernhardmgruber: That's the general bummer with all those C++ version annotations on the slides. They currently mean what C++ version you need to compile the slide content.

I kind of also want to avoid writing `std::is_floating_point<T>::value`, because students should definitely use `std::is_floating_point_v<T>`. It's a dilemma.

talk/basicconcepts/control.tex
bernhardmgruber: That optimization is indeed surprising :D If you compare just the disassembly for the implementation of an `is_even` (https://godbolt.org/z/968nEeqsq), which is what the condition of the ternary is, the compiler generates the same code.

The main point for me is to not encourage using a bitwise op over a modulo when testing whether an integral is even. Could you afford a linebreak to keep the better spacing and leave the `a % 2 == 0`?

talk/basicconcepts/functions.tex
bernhardmgruber: Also keep in mind that there is no single ABI :) ARM and RISC-V are probably going to have their own rules for what is passed in registers and what on the stack. Windows also has it's own ABI and you can alter that with various calling conventions (e.g. `__vectorcall` and `__fastcall`). We should definitely not specify a size. I also tend to avoid pointing at a concrete ABI specification.

Btw, how is your ABI spec different than the Itanium C++ ABI here: https://github.com/itanium-cxx-abi/cxx-abi/? I thought the latter was the one that gcc/clang implement for Linux.

talk/basicconcepts/references.tex
bernhardmgruber: ok. no further change necessary!

talk/basicconcepts/control.tex
amadio: I don't see a problem with the bitwise operator, modulo is really slow if one is not careful. I think I will just drop this commit from my pull request.

talk/basicconcepts/assert.tex
amadio: I think it's preferable to use `::value` so students don't get a wrong impression that `static_assert` is a C++17 feature. I got that impression the first time I saw this slide.

talk/basicconcepts/control.tex
bernhardmgruber: You are absolutely right! I am just talking about the case to express whether an integral is even, where I prefer the modulo for readability. Maybe others want to share their opinion? @hageboeck ?

talk/basicconcepts/assert.tex
bernhardmgruber: ok, then go for it!

talk/basicconcepts/functions.tex
amadio: I can remove the link, but amd64/x86_64 is pretty much the only thing people in the course would be using. The Itanium C++ ABI details only the C++ specific parts, it is built on top of the generic amd64/x86_64 ABI, which is not specific to C++ (see http://itanium-cxx-abi.github.io/cxx-abi/abi.html#docs). What do you recommend to say, then, if you don't want to give a specific size and don't want to use the pointer size (i.e. reference size) as a loose guideline?

talk/basicconcepts/control.tex
hageboeck: In terms of readability:
Don't use collatz at all. Something *much simpler* might illustrate the usage better than something that you have to parse recursively in your head.

talk/basicconcepts/functions.tex
hageboeck: Remember that this is for the essentials course. The guideline needs to be very simple.
"numbers" is indeed a bit fuzzy, but something that is "a few bytes" or "about the size of a pointer" (== a few bytes) or "a primitive type" or a small struct should all be fine.

talk/basicconcepts/control.tex
hageboeck: e.g.
```c++
std::cout << "The condition is " << (condition ? "true" : "false")
```

talk/basicconcepts/functions.tex
hageboeck: If you want the "bigger" to be honoured, say "a few times the size of a pointer"

talk/basicconcepts/control.tex
amadio: I agree, but this is out of scope for my changes, I will just drop the commit so we can proceed to merge the other changes.

talk/basicconcepts/functions.tex
amadio: I've dropped this commit from the changeset, let's address this separately.

bernhardmgruber (approved):

amadio:
Thanks! I will submit the next changes soon.

sponce:
I'm realizing this was merged only now as... nothing works anymore on my local box !
You're supposing here a number of things which are non standard, starting with the presence of latekmk on the system, which I did not have. Ok, installing was not too difficult, but now it fails at least in 2 ways : 
  - make clean does not do much (you've dropped all the lines dropping temporary files)
  - make fails, complaining on missing  File `fontawesome.sty'

Can we go back to more standard tools ?

sponce:
In case others have the same issue, package texlive-fonts-extra brings fontawesome.sty.

Except installing it depended on another 41 packages and used 1.6GB of disk !!!
Can we go back to more lightweight things ?

Also I found another issue : once the build failed, installing the missing pieces does not help. The build will fail again and again if you do not do a manual `make clean`

Last point : the build is now completely silent. You have no clue what's going on. I admit that the previous output was a bit too generous, but I'd like to see some progress. (Side note, we should fix the problems leading to the generosity of the previous output, these were all real warnings)

amadio:
Hi @sponce. I understand your frustration, but let me explain things a bit so you can decide if you want to revert changes, or if you can just adjust what's there. I based the changes on what was available in the image, as I thought that most people were using GitHub to build and only downloading the PDF later (I saw problems locally when trying to build myself the first time I tried to build). Latexmk is quite common to use for projects that have to be built multiple times, as it takes care of rebuilding when necessary, calling bibtex, etc.

Let's start with `make clean`. There were two targets, `clean` and `clobber`, which I mapped to `latexmk -c` and `latexmk -C`, the second one should leave you only the PDF, while the first will clean only some files to allow for a faster recompilation. The file `talk/.latexmkrc` has settings to control what's removed:

```
$clean_ext = "cut lex lgb lgp mw nav snm vrb _minted-%R/* _minted-%R";
```

Try `make clobber` and let me know if that's not doing the job for you. You can add more extensions to be removed to `clean_ext`, or we can just make `make clean` call what's now called by `make clobber` and drop `make clobber`.

The `fontawesome.sty` missing is not something related to the build changes, but to the YouTube channels page, where I added symbols based on this font. If you think it brings too many dependencies, then we'd have to replace the calls to fontawesome with images. Locally, you can just comment out that slide if you don't want to install it.

The build was extremely verbose before, but if you don't like how it's now, it's just a matter of removing the `-quiet` option being passed to latexmk in the `Makefile` if you want to make it permanent, or using `make V=1` to have the verbose output enabled.

sponce:
> I thought that most people were using GitHub to build 

For tiny fixes maybe (although I do not) but for developing new set of slides, it's not really working. Delay is far too long and you cannot compile only your chapter.

> Let's start with make clean. There were two targets, clean and clobber

Ok, here my problem seems to be lying in the fact that you've changed the meaning of clean and clobber. What used to be clean is now clobber (leaving only the pdf) and clobber does not exist anymore (was dropping the pdf on top of clean). But you've introduced a "clean only some files to allow for a faster recompilation" which was not existing. I'm not sure whether that last one is useful/safe, but certainly I was caught by the change of meaning of the original ones

> The fontawesome.sty missing is not something related to the build changes, but to the YouTube channels page,

Why did we need that special font there ? I would indeed avoid bringing more dependencies than needed if it does not bring a real plus. here I do not see why youtube page would not use the same font as the rest.

> The build was extremely verbose before

Yes indeed, but all lines were actual issues. So we should rather solve them than hide them. Of course I've never manage to find time for it and we all know how picky latex can be in these cases. But my fear is that we know hide more serious stuff.

Bottom line : I believe we can have a nice working state by : 
  - changing the font for youtube
  - removing the -quiet in the Makefile
  - fixing the make clean. I believe my problem was that .log file should have been dropped also. Not sure
Can you try it ?

amadio:
Yeah, only the page you see on #383 needs the font, to show the YouTube channel symbol before the channel names. I can replace that with a small SVG or PNG file.

I can remove `-quiet` if you don't like to just type `make V=1`, I think I will make verbose the default, and add `make QUIET=1` to have a quiet build then, or just override the latexmk options adding `-quiet` by hand when necessary.

I will make `make clean` remove all but the PDF, and `make clobber` remove all including the PDF. I will put up a pull request for this soon.

sponce:
Thanks a lot. For the youtube part, can we not just drop the symbol and replace the red part by "youtube" on each line. I mean, we just have twice the same words there.
By the way, maybe we should darken the red to make it more readable

amadio:
One curiosity, after `dnf install texlive`, a call to `dnf install texlive-fontawesome` just added a few kb to my system:
```
thinkpad talk $ sudo dnf install texlive-fontawesome
Place your finger on the fingerprint reader
Failed to match fingerprint
Place your finger on the fingerprint reader
Failed to match fingerprint
Place your finger on the fingerprint reader
Failed to match fingerprint
[sudo] password for amadio: 
Last metadata expiration check: 1:54:02 ago on Wed 25 Jan 2023 09:43:34 AM CET.
Dependencies resolved.
===============================================================================================================================================================================================================================================================================================================================
 Package                                                                             Architecture                                                           Version                                                                            Repository                                                                 Size
===============================================================================================================================================================================================================================================================================================================================
Installing:
 texlive-fontawesome                                                                 noarch                                                                 9:20200406-25.el9                                                                  appstream                                                                 519 k

Transaction Summary
===============================================================================================================================================================================================================================================================================================================================
Install  1 Package

Total download size: 519 k
Installed size: 699 k
Is this ok [y/N]: y
Downloading Packages:
texlive-fontawesome-20200406-25.el9.noarch.rpm                                                                                                                                                                                                                                                 6.4 MB/s | 519 kB     00:00    
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Total                                                                                                                                                                                                                                                                                          6.3 MB/s | 519 kB     00:00     
Running transaction check
Transaction check succeeded.
Running transaction test
Transaction test succeeded.
Running transaction
  Preparing        :                                                                                                                                                                                                                                                                                                       1/1 
  Installing       : texlive-fontawesome-9:20200406-25.el9.noarch                                                                                                                                                                                                                                                          1/1 
  Running scriptlet: texlive-fontawesome-9:20200406-25.el9.noarch                                                                                                                                                                                                                                                          1/1 
  Verifying        : texlive-fontawesome-9:20200406-25.el9.noarch                                                                                                                                                                                                                                                          1/1 

Installed:
  texlive-fontawesome-9:20200406-25.el9.noarch
```
What distro are you using?

amadio:
> Thanks a lot. For the youtube part, can we not just drop the symbol and replace the red part by "youtube" on each line. I mean, we just have twice the same words there.
> By the way, maybe we should darken the red to make it more readable

Using "YouTube" for all the channels will likely look worse than now, and CppCon's YouTube channel is called BoostCon, and other channels have variations too. I think it will look better if we just use an image for the channel icon. I will darken the red a bit, though, as you suggested.

sponce:
> What distro are you using?

debian unstable. But it of course all depends on what was already installed. I have a quite small system, built from scratch with only what I need.

For YouTube, ok for me the way you describe it

PR #389, opened 2023-01-09
Disable automatic sync of stale bot config

klieret:


hageboeck (commented):
Unless this PR is still work in progress, I don't see how removing 8 lines of comments should have an effect.
Could you explain a bit how this would affect the sync?

klieret:
It's because I use [this script](https://github.com/hsf-training/maintenance/blob/main/add-common-files/add-files.py) to add the files and it will look for the markers in the comment.

hageboeck (approved):
OK, got it. Thanks!

PR #395, opened 2023-02-13
Simplify move exercise.

### LINKED ISSUE #340: Move exercise too complicated
Comment by a participant: "You can't see the wood for the tree"

There's more templates and complications than needs to be to demonstrate moving.

hageboeck:
For many students, it took too long to read the NVector implementation. In this version, we use a class with one pointer to internal storage to compare copy and move.

Fix #340.


code/move/solution/trymove.sol.cpp
```diff
@@ -1,39 +1,87 @@
...
-        int a = d(e);
-        int b = d(e);
-        swap(v, a, b);
+/**
+ * An attempt to make a movable class. It allocates memory on construction, and has
+ * array semantics. (Prefer std::array or std::vector instead of hand-crafted solutions, though.)
+ * We made it copyable by providing copy constructor and assignment operator, but note that we
+ * are violating the rule of 5, so it's not moveable yet.
+*/
+class Moveable {
```
bernhardmgruber, line 11: Could we give this a more descriptive name? Like e.g. `DynArray`.

code/move/solution/trymove.sol.cpp
```diff
@@ -1,39 +1,87 @@
...
+    static constexpr std::size_t size = 10000;
+
+public:
+    Moveable() = default;
+    Moveable(Moveable const & other) :
+        Moveable()
+    {
```
bernhardmgruber, line 18: Why did you delegate to the default ctor?
```suggestion
    Moveable(Moveable const & other) :
    {
```

code/move/solution/trymove.sol.cpp
```diff
@@ -1,39 +1,87 @@
...
+        // Delegate copying to the assignment operator
+        *this = other;
     }
-}
+    Moveable & operator=(const Moveable & other)
+    {
+        if (other.m_storage != m_storage) {
```
bernhardmgruber, line 24: This is protection against self-assignment, is it? Then why are you not using the canonical form?
```suggestion
        if (&other != this) {
```

code/move/solution/trymove.sol.cpp
```diff
@@ -1,39 +1,87 @@
...
+    // Note how we are avoiding the initial allocation by directly
+    // move-initialising the pointer
+    Moveable(Moveable && other) noexcept :
+        m_storage{ std::move(other.m_storage) }
+    {  }
+    Moveable & operator=(Moveable && other) noexcept
+    {
+        if (m_storage != other.m_storage) m_storage = std::move(other.m_storage);
+        return *this;
+    }
```
bernhardmgruber, line 43: I usually implement move ctor and move assign by calling a `void swap(Moveable&)` member function. I don't recall what we discuss in the slides.

code/move/solution/trymove.sol.cpp
```diff
@@ -1,39 +1,87 @@
...
+
+    int & operator[](std::size_t index) { return m_storage[index]; }
+
+private:
+    std::unique_ptr<int[]> m_storage { new int[size] };
```
bernhardmgruber, line 48: For me, naked `new` is a code smell, so I would use here:
```suggestion
    std::unique_ptr<int[]> m_storage = std::make_unique<int[]>(size);
```

bernhardmgruber (commented):
I like the simplification, thx! Here are some comments.

code/move/solution/trymove.sol.cpp
hageboeck: I wanted to call it something with array first, but at the same time that might connect this things with real arrays in the minds of the students. We don't want that, do we?

code/move/solution/trymove.sol.cpp
hageboeck: Because initially, I used it to allocate. It's not needed, any more, but is there a reason not to delegate?

code/move/solution/trymove.sol.cpp
bernhardmgruber: But your type is kind of array-ish, it has a size and a subscript :)

code/move/solution/trymove.sol.cpp
hageboeck: I'm not aware of a canonical form.

code/move/solution/trymove.sol.cpp
bernhardmgruber: Well, it's dead code, but otherwise there is no reason. Keep it if you want.

code/move/solution/trymove.sol.cpp
hageboeck: I would have to upgrade to c++17 to do this. I wasn't sure if that should be an option.

code/move/solution/trymove.sol.cpp
bernhardmgruber: See e.g. here: https://en.cppreference.com/w/cpp/language/operators#Assignment_operator. You can guard against self-assignment by comparing `other`'s identity with `this`.

code/move/solution/trymove.sol.cpp
bernhardmgruber: C++17 is the minimum for all exercises already, see e.g. here: https://github.com/hsf-training/cpluspluscourse/pull/356#issuecomment-1276706146

code/move/solution/trymove.sol.cpp
hageboeck: Copy and swap, I'll put that in the solution.

code/move/solution/trymove.sol.cpp
hageboeck: I call it CustomArray now, to give it a bad connotation.

bernhardmgruber (approved):

PR #396, opened 2023-02-14
Improve race exercise.

### LINKED ISSUE #353: Atomic exercise DOES NOT race on my machine
@hageboeck @sponce @bernhardmgruber 

I suspect the `-O2` compiling option (with g++ 11.3. ) make that `for (int i = 0; i < 100; ++i) { a++; }` is some howreplaced with `a+=100` ... Quite simple to change in `Makefile`. **How to do it in `CMakeLists.txt` ?**

I also needed to move up the code to 10000 tries (instead of 1000), and the main function to 1000 increments (instead of 100) before I start to see racing conditions.

hageboeck:
The race exercise races very infrequently on some platforms. Therefore:
- Provide a variable to spawn more threads
- Provide a run script that invokes the executable until a race occurs
- Add thread sanitizer and valgrind as optional steps to the instructions

Fix #353.

### DIFF
--- a/code/race/racing.cpp
+++ b/code/race/racing.cpp
@@ -1,10 +1,16 @@
 #include <iostream>
 #include <thread>
+#include <vector>
 
 /*
- * This program tries to increment an integer 200 times in two threads.
- * Check whether the result is indeed always 200.
+ * This program tries to increment an integer 100 times from multiple threads.
+ * If the result comes out at 100*nThread, it stays silent, but it will print
+ * an error if a race condition is detected.
+ * If you don't see it racing, try ./run ./racing, which keeps invoking the
+ * executable until a race condition is detected.
  */
+
+constexpr unsigned int nThread = 2;
 
 int main() {
   int nError = 0;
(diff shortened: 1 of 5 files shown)


code/race/racing.cpp
```diff
@@ -19,20 +25,19 @@ int main() {
...
-    // Run with two threads
-    std::thread t1(inc100);
-    std::thread t2(inc100);
-    for (auto t : {&t1,&t2}) t->join();
+    // Start up all threads:
+    std::vector<std::thread> threads;
+    for (unsigned int i = 0; i < nThread; ++i) {
+      threads.emplace_back(inc100);
+    }
+    for (auto & thread : threads) thread.join();
```
bernhardmgruber, line 33: Haha, I looked at your code and thought: This is just:
```suggestion
    std::vector<std::jthread>(nThread, inc100);
```
But I am not sure if this is more readable :D

bernhardmgruber (approved):

code/race/racing.cpp
hageboeck: It doesn't compile AFAIK, since you have to pass a thread and not the lambda to the constructor.
Be my guest:
https://godbolt.org/z/qKM76rosE

hageboeck: Or phrased differently: I don't know a way to fill the vector other than emplacing or moving the threads.

bernhardmgruber: Right, vector's ctor expects its T as the ctor argument, and not a type convertible to T. Got me :)

PR #397, opened 2023-02-14
Remove C-style arrays from smart pointers exercise.

### LINKED ISSUE #163: Use std::vector in smartPointers and stl exercises ?
It is my strong feeling that the use of raw C arrays makes the syntax uselessly complex, and divert the learners from what we want to demonstrate. It would rather use std::vector, provided it has been previously presented in the morning lessons.

hageboeck:
Fix #163.

### DIFF
--- a/code/smartPointers/Makefile
+++ b/code/smartPointers/Makefile
@@ -5,7 +5,7 @@ clean:
 	rm -f *o *so smartPointers *~ smartPointers.sol
 
 % : %.cpp
-	$(CXX) -g -std=c++17 -Wall -Wextra -o $@ $<
+	$(CXX) -g -std=c++20 -Wall -Wextra -o $@ $<
 
 %.sol : solution/%.sol.cpp
-	$(CXX) -g -std=c++17 -Wall -Wextra -o $@ $<
+	$(CXX) -g -std=c++20 -Wall -Wextra -o $@ $<
--- a/code/smartPointers/smartPointers.cpp
+++ b/code/smartPointers/smartPointers.cpp
@@ -14,9 +14,11 @@
  * Please fix all memory leaks / ownership problems using smart pointers.
  * (Verify by running the program with valgrind!)
  *
+ * In main() at the bottom, comment in all parts as you progress through the exercise.
+ *
  * Remember that:
  * - The unique ownership of data is expressed using unique_ptr.
- * - "Observer" access without ownership is expressed using raw pointers or references.
+ * - "Observer" access without ownership is expressed using raw pointers, references or spans.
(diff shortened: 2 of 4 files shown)


code/smartPointers/smartPointers.cpp
```diff
@@ -30,21 +31,21 @@
  */
 
 // Declare a function to do something with the data. No need to change it.
-double sumEntries(const double* data, std::size_t size) {
-    if (size > 200)
-        throw std::invalid_argument("I only want to sum 200 numbers or less.");
+template<typename T>
+double sumEntries(const T * data) {
```
bernhardmgruber, line 35: I was sure there is an error when you call `data->begin()` later until I realized you pass a `std::array<T>*`. Consider renaming the template paramter to make this clearer:
```suggestion
template<typename Range>
double sumEntries(const Range * data) {
```

hageboeck: > I was sure there is an error when you call data->begin()

I compile before I commit. 🙂
But it's a good idea to call it range, so we'll go for that!

code/smartPointers/smartPointers.cpp
```diff
@@ -30,21 +31,21 @@
  */
 
 // Declare a function to do something with the data. No need to change it.
-double sumEntries(const double* data, std::size_t size) {
-    if (size > 200)
-        throw std::invalid_argument("I only want to sum 200 numbers or less.");
+template<typename Range>
+double sumEntries(const Range * data) {
```
sponce, line 35: Maybe let's be slightly simpler by avoiding the template and using std::span ?

sponce (approved):

chavid (approved):
Ok for me.

sponce:
somehow checks are failing because we try to compile for C++11 something that clearly needs C++20. Not sure what is wrong

code/smartPointers/Makefile
```diff
@@ -5,7 +5,7 @@ clean:
 	rm -f *o *so smartPointers *~ smartPointers.sol
 
 % : %.cpp
-	$(CXX) -g -std=c++17 -Wall -Wextra -o $@ $<
+	$(CXX) -g -std=c++20 -Wall -Wextra -o $@ $<
 
 %.sol : solution/%.sol.cpp
-	$(CXX) -g -std=c++17 -Wall -Wextra -o $@ $<
+	$(CXX) -g -std=c++20 -Wall -Wextra -o $@ $<
```
bernhardmgruber, line 11: You may want to apply the same changes to `CMakeLists.txt` as well.

hageboeck: Done, all green. Merging now.

PR #402, opened 2023-02-22
Better example for usage of enable_if

### LINKED ISSUE #378: 5th course, notes
Just few comments I noted down from the last course

slide 333

	change f(f())

slide 350:

	change the “Gaudi example” slide on enable_if_t, there is a better way, which?

sponce:
Closes #378

### DIFF
--- a/talk/expert/sfinae.tex
+++ b/talk/expert/sfinae.tex
@@ -236,22 +236,28 @@
 \end{frame}
 
 \begin{frame}[fragile]
-  \begin{exampleblock}{Gaudi usage example}
+  \begin{exampleblockGB}{Usage example}{https://godbolt.org/z/xse8GM8fq}{\texttt{enable\_if}}
+    \small
     \begin{cppcode*}{}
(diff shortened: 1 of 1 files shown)


talk/expert/sfinae.tex
```diff
@@ -236,22 +236,29 @@
...
+        template<typename T,
+                 typename std::enable_if_t
+                   <is_iterable<T>, bool> = true>
+        void print(T const& t) {
+            for (auto const& item : t) {
+                std::cout << item << "\n";
+            }
         }
-      } deref;
+    };
```
bernhardmgruber, line 259: I would remove the outer struct `Printer`. And to implement such a function `print` I would use `if constexpr` here :)

sponce: Indeed. So I've remove the struct.

For the if constexpr, yes, indeed and that's the point of a further slide. I've added a note at the bottom though and implemented both solutions in the godbolt example.

Issue #399, opened 2023-02-15
Better example for function objects


bernhardmgruber:
Our current example for function objects is:
```c++
    struct BinaryFunction {
      virtual double operator() (double a, double b) = 0;
    };
    double binary_op(double a, double b, BinaryFunction &func) {
      return func(a, b);
    }
    struct Add : BinaryFunction {
      double operator() (double a, double b) override
      { return a+b; }
    };
    Add addfunc;
    double c = binary_op(a, b, addfunc);
```
which I think is really bad, since function objects are used with templates since the beginning of the STL. Furthermore, I don't want to encourage the use of inheritance for function objects. We should come up with a better slide here.

PR #404, opened 2023-02-22
Replace functor example using inheritance

bernhardmgruber:
Fixes: #399

### DIFF
--- a/talk/objectorientation/functors.tex
+++ b/talk/objectorientation/functors.tex
@@ -7,7 +7,7 @@
     \item also known as functors (no relation to functors in math)
     \item a class that implements \mintinline{cpp}{operator()}
     \item allows to use objects in place of functions
-    \item and, as objects have constructors, allow to construct functions
+    \item with constructors and data members
     \end{itemize}
   \end{block}
   \begin{cppcode}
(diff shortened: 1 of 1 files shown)


talk/objectorientation/functors.tex
```diff
@@ -25,25 +25,28 @@
...
     \end{itemize}
   \end{block}
+  \small
   \begin{cppcode}
-    struct BinaryFunction {
-      virtual double operator() (double a, double b) = 0;
-    };
-    double binary_op(double a, double b, BinaryFunction &func) {
-      return func(a, b);
+    int count_if(const auto& range, auto predicate) {
```
bernhardmgruber, line 36: To show a higher-order function here, we need templates. I chose to use `auto` here, since we already discussed the `auto` keyword and it should naturally follow what's going on here.

sponce: The example is indeed better now, but it requires C++20 and makes an expert use of auto, which was not yet explained (and is not at all I believe). That was precisely the reason of being of the original base class !

To word it differently, you're replacing inheritance by templating. It's almost always a good move, expect we did not yet introduce templating at that stage.

Not sure what the conclusion is. @hageboeck any opinion ?

hageboeck: Hmm, I agree that it goes two steps ahead. Maybe we should point out that it's a template, and that these are going to be explained later?

hageboeck: Thinking about it, I don't see how to do it without templates or inheritance. Maybe we should indeed show "the better way" of doing it, give a brief explanation (verbally) what's going on, and refer to the template chapter on the slide.

sponce: Ok, let's go for that. Last detail, @bernhardmgruber can you add a godbolt URL ?

sponce (approved):

Issue #256, opened 2022-08-24
Consider removing second python example in lambda section


bernhardmgruber:
I'm fine with the first Python analogy. That's indeed helpful for people who know it already. But the second one about capture only shows that there's nothing specific to be done in Python. Here I don't see how it helps, because the message is:
In C++ it's different than in Python.

_Originally posted by @hageboeck in https://github.com/hsf-training/cpluspluscourse/pull/250#discussion_r953685189_

bernhardmgruber:
This is about this slide:
![image](https://user-images.githubusercontent.com/1224051/186419244-98061574-f31c-4c39-94c2-652594c00d70.png)

hageboeck:
Instead, one could use the space of the first box for some kind of explanation like:
- Task: Write a functor to increment values in a collection
- How do you avoid writing one lambda for every possible value of `increment`?
- `increment` needs to be "captured"

PR #405, opened 2023-02-22
Replace second python lambda by text

bernhardmgruber:
This PR tries to improve what's described in issue #256. However, I am not sure whether this is actually an improvement. I would also be happy with the status quo. @hageboeck you asked for it, your opinion please :)

### DIFF
--- a/talk/morelanguage/lambda.tex
+++ b/talk/morelanguage/lambda.tex
@@ -75,12 +75,11 @@
 
 \begin{frame}[fragile]
   \frametitlecpp[11]{Capturing variables}
-  \begin{block}{Python code}
-    \begin{pythoncode*}{}
-      increment = 3
-      data = [1,9,3,8,3,7,4,6,5]
-      map(lambda x : x + increment, data)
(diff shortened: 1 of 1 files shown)


talk/morelanguage/lambda.tex
```diff
@@ -75,12 +75,11 @@
...
-  \begin{block}{Python code}
-    \begin{pythoncode*}{}
-      increment = 3
-      data = [1,9,3,8,3,7,4,6,5]
-      map(lambda x : x + increment, data)
-    \end{pythoncode*}
+  \begin{block}{Adaptable lambdas}
+    \begin{itemize}
+      \item we can adapt a lambda's behavior
+      \item by accessing variables outside its body
```
hageboeck, line 81:
```suggestion
      \item We can adapt a lambda's behaviour by accessing variables outside its body
      \item This is called ``capture''
```

hageboeck (approved):
Here's a suggestion for initial text, but otherwise LGTM.

Issue #394, opened 2023-02-06
Beginner Course - Debugging


TobiasDuswald:
Dear all,

I wanted to follow up on a brief conversation I had with @roiser regarding debugging tools. First of all, I think it's great that it's part of your beginner track because I often see this topic being ignored while, in fact, it's one of the most important topics there is. I reached out to @roiser to "advertise" the graphical debuggers. It seems like over time people fell back to using gdb / lldb via the terminal, but there are arguably many advantages to using GUIs for debugging. E.g., [CodeLLDB](https://marketplace.visualstudio.com/items?itemName=vadimcn.vscode-lldb) is a free extension for VSCode integrating debugging of CPP code in the IDE. There are many alternatives (see for instance [ATPESC2022](https://extremecomputingtraining.anl.gov/agenda-2022/), Track 6) but some of them are not free.

Point of this issue is that I think it would be very helpful to make people aware of these tools such that they at least know they exist and can follow up if they wanted to. So I think it would be good to consider allocating 1-2 slides to expose participants to these tools. :) 

Tobias

bernhardmgruber:
Hi, thank you for your suggestion!

> It seems like over time people fell back to using gdb / lldb via the terminal, but there are arguably many advantages to using GUIs for debugging.

In my experience this depends a lot on your organisation's culture. For many software engineers or folks in the industry using a graphical IDE is a no-brainer. Others just start out with the bare and established command line tools and manage to pass the long learning curve to also reach stunning productivity. It is ultimately a mix of culture, education, work environment and personal preference.

> [...] but some of them are not free.

Which is often a big driver for choosing a specific solution. A few hundret bucks sounds expensive, especially when you are a student or in your early career. However, when you start to think about the value of your own time, a good development environment can already amortize its cost at the first bug that took you only half the time to find :)

CERN has licenses for some products, like MS Visual Studio which offers a great experience on Windows. Your question now motivated me to inquire at the license service whether CERN could get licenses for CLion, which I have been using for 2 years now on Linux and also offers a great experience.

> Point of this issue is that I think it would be very helpful to make people aware of these tools such that they at least know they exist and can follow up if they wanted to. So I think it would be good to consider allocating 1-2 slides to expose participants to these tools. :)

[...]

PR #406, opened 2023-02-22
Add a slide on debug UIs

bernhardmgruber:
I wonder whether we should include some screenshots of some tools, so people can have a quick feel of what they would get.

Fixes: #394


sponce (approved):

sponce:
I would avoid images in these slides when not needed, just so that we have a nice, easy to use 5MB file rather than an enormous 250MB file due to all the images. Btw, current images are already ~2MB out of the 5 !

PR #407, opened 2023-02-23
Fixup for #405

bernhardmgruber:
This PR is a fixup for PR #405, which was merged before I could push :)


sponce:
So now I do not dare merging this one :-D

PR #393, opened 2023-01-30
Add a slide on weak_ptr.

hageboeck:
First draft for explaining a weak_ptr.
I'm still thinking whether a cache is the best way to explain a `weak_ptr`. It might draw people into a direction where thread safety is an issue.

Fix #347.

PR #410, opened 2023-02-23
Move weak_ptr slides into advanced course.

hageboeck:


talk/morelanguage/raii.tex
```diff
@@ -318,9 +320,11 @@
   \end{exampleblock}
 \end{frame}
 
+\end{advanced}
+
 \begin{frame}[fragile]
-  \frametitlecpp[11]{Quiz: shared\_ptr and weak\_ptr}
-  \begin{exampleblockGB}{What is the output of this code?}{https://godbolt.org/z/vM35Y6qEW}{shared\_ptr vs weak\_ptr}
+  \frametitlecpp[11]{Quiz: shared\_ptr in use}
+  \begin{exampleblockGB}{What is the output of this code?}{https://godbolt.org/z/vM35Y6qEW}{shared\_ptr quiz}
```
bernhardmgruber, line 327: As pointed out in my post-merge feedback to #393, I think we typeset C++ constructs in headers using `\texttt{...}`:
```suggestion
  \frametitlecpp[11]{Quiz: \texttt{shared\_ptr} in use}
  \begin{exampleblockGB}{What is the output of this code?}{https://godbolt.org/z/vM35Y6qEW}{\texttt{shared\_ptr} quiz}
```

bernhardmgruber (approved):
I also think `weak_ptr` should be rather an advanced topic, since they are used less often.

talk/morelanguage/raii.tex
hageboeck: At best, we are inconsistent:
image
image
image
image
image
image

I agree it should be discussed. Let's also ask @sponce 

In this specific case, however, (line 327), I'm not even sure that a typewriter font is a good choice. The last argument is what ends up in this list:
image

Here's a proposal, but feel free to chime in:
| Environment | What to do |
| --- | --- |
| `\frametitlecpp` | `\texttt{ }` |
| Chapter/section title | `\texttt{ }` |
| short chapter/section title | normal font? `\texttt{ }`?? (might break when linking / indexing) |
| Box title | `\mintinline{cpp}{ ... }` |
| Inline text | `\mintinline{cpp}{ ... }` |
| Index of XXX (like image above) | normal font? |

bernhardmgruber: Ah ok, so maybe use a normal font for the index then. The section titles may be normal as well to fit on the overview slide. Otherwise, I would use `\texttt{}` whereever we can. And yes, that is currently very inconsistent.

sponce: I would respect the simple rule "always use \mintinline unless the context does not allow, then go back to \texttt". But basically it matches what @hageboeck proposes, answering no to the 2 questions about normal font.

hageboeck: Now done. I'll merge when the CI is green.

PS @sponce @bernhardmgruber:
I checked if `\mintinline{cpp}` works in frame titles. When it finds something to highlight, it gets hard to read.

PR #414, opened 2023-03-01
Use accumulate over reduce when necessary

bernhardmgruber:
`std::reduce` can only be used over `std::accumulate` when the binary operation is commutative and associative. This does not hold for the exercises computing squared sums, so `std::reduce` is a bug there.

### DIFF
--- a/code/header_units/main.cpp
+++ b/code/header_units/main.cpp
@@ -24,9 +24,9 @@ void compute(int len, T initial, T step) {
     std::adjacent_difference(v.begin(), v.end(), diffs.begin());
 
     // compute standard deviation of it
-    const T sum = std::reduce(diffs.begin()+1, diffs.end(), T());
-    const T sumsq = std::reduce(diffs.begin()+1, diffs.end(), T(),
-                                [](const T& s, const T& a) { return s + a * a; });
+    const T sum = std::reduce(diffs.begin()+1, diffs.end());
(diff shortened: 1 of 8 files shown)


hageboeck:
I'm puzzled by the motivation for the change. The op should be commutative, but it's likely not associative (i.e. different groupings lead to a different end result). I wouldn't say is wrong, though. It's not deterministic. Did I overlook something?

hageboeck:
BTW: The same would hold for the simple `sum`.

bernhardmgruber:
> The op should be commutative, but it's likely not associative (i.e. different groupings lead to a different end result).

We are both right. The [documenation ](https://en.cppreference.com/w/cpp/algorithm/reduce) says for the binary function:

`The behavior is non-deterministic if binary_op is not associative or not commutative.`

`auto op = [](const T& s, const T& a) { return s + a * a; }` is definitely neither commutative nor associative. `op(1, 2) == 5` and `op(2, 1) == 3`. and `op(1, op(2, 3)) == 122` and `op(op(1, 2), 3) == 14`.

> I wouldn't say is wrong, though. It's not deterministic. Did I overlook something?

For computing the sum of squares, this is definitely wrong IMO. For the plain sum, your may get different rouding errors, which are fine IMO.

hageboeck:
Oh, I didn't see that the op takes only one square ...
It's better to change this, indeed.

PR #418, opened 2023-03-03
Advertise lxplus9

bernhardmgruber:
All the tools (except for GUI's like `kcachegrind`) should be available there, even without using an LCG view.


hageboeck (approved):
LGTM!

PR #420, opened 2023-03-05
Remove alert and qualify fixed-width int types

bernhardmgruber:
Not urgent and up for debate.

I would like to replace the "aggressive" red alert boxes by simple comments in the code snippets. Furthermore, all the types mentioned which come from C++ standard headers actually require to be prefixed `std::`. Most compilers allow their use without the namespace though. One argument for the prefix is that such code stops compiling when the `#include <cstdint>` is replaced by `#import std;` (which is why there is `#import std.compat`, preserving the old behavior.).

### DIFF
--- a/talk/basicconcepts/coresyntax.tex
+++ b/talk/basicconcepts/coresyntax.tex
@@ -81,22 +81,21 @@
 
 \begin{frame}[fragile]
   \frametitlecpp[98]{Portable numeric types}
-  \alert{Requires inclusion of a specific header}
   \begin{cppcode}
-    #include <cstdint>
+    #include <cstdint> // defines the following:
 
-    int8_t c = -3;     // 8 bit signed integer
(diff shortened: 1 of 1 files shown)


talk/basicconcepts/coresyntax.tex
```diff
@@ -144,20 +143,19 @@
...
-    size_t s = sizeof(int); // unsigned integer
+    std::size_t s = sizeof(int); // unsigned integer
                             // can hold any variable's size
 
-    #include <cstdint>
+    #include <cstdint> // defines these:
 
-    ptrdiff_t c = &s - &s;  // signed integer, can hold any
+    std::ptrdiff_t c = &s - &s;  // signed integer, can hold any
                             // diff between two pointers
```
sponce, line 155: Does not fit anymore in the line.
The comment should be put before the code

bernhardmgruber: done

Issue #213, opened 2022-08-11
C++23: Add slide on extended floating point types


bernhardmgruber:
C++23 will bring the following extended floating point types:

std::float16_t
std::float32_t
std::float64_t
std::float128_t
std::bfloat16_t

Notably, support for 16 and 128bit floats.
bfloat16 is used by Google's TPUs, in Tensorflow and on NVIDIA hardware.

See also https://wg21.link/p1467

bernhardmgruber:
The floating-point literal slide in the basics course alreay mentions a few:
![image](https://user-images.githubusercontent.com/1224051/222968786-31663ea3-24ef-4cb3-a0dd-329e029c936d.png)

I think we should move those literals on a dedicated slide together with the new extended FP types as discussed above.

PR #422, opened 2023-03-05
Add a slide on extended floating-point types

bernhardmgruber:
No need to rush this in. I can also explain the situation during the course tomorrow.

This also fixes a small error on the FP slides, where we claim that `3.14F32` is a `float`, which if I read https://wg21.link/P1467 again, is wrong. It is a `std::float32` which is a type distinct from `float`. g++ trunk confirms this: https://godbolt.org/z/TE96bzK7W.

However, the way this works changed a few times during standardization, so I blame noone with getting this wrong :D

Fixes: #213

### DIFF
--- a/talk/basicconcepts/coresyntax.tex
+++ b/talk/basicconcepts/coresyntax.tex
@@ -73,8 +73,8 @@
     long long ll = 0LL;          // min 64 bit
     unsigned long long l = 0ULL; // min 64 bit
 
-    float f = 1.23f;          // 32 (23+8+1) bit float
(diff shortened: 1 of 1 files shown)


talk/basicconcepts/coresyntax.tex
```diff
@@ -161,3 +161,26 @@
     uintptr_t i = reinterpret_cast<uintptr_t>(&s); // unsigned
     \end{cppcode}
 \end{frame}
+
+\begin{frame}[fragile]
```
sponce, line 165: Let's mark this one advanced

PR #429, opened 2023-03-06
Small exercise fixes

bernhardmgruber:
Here are a few small fixes that I collected during the exercises.

### DIFF
--- a/code/control/control.cpp
+++ b/code/control/control.cpp
@@ -5,8 +5,8 @@ constexpr unsigned int numbers[]{1000001, 1000002, 1000003, 1000004, 1000005, 10
 bool isodd(unsigned int i) { return i % 2 == 1; }
 
 void part1() {
-  int sum_odd = 0;
-  int sum_eve = 0;
+  unsigned int sum_odd = 0;
+  unsigned int sum_eve = 0;
   for (int i = 0; i < 9; ++i) {
     unsigned int num = numbers[i];
     if (isodd(num)) {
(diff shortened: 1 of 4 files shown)


code/control/control.cpp
```diff
@@ -1,12 +1,12 @@
 #include <iostream>
 
-constexpr unsigned int numbers[]{1000001, 1000002, 1000003, 1000004, 1000005, 1000006, 1000007, 1000008, 1000009};
+const unsigned int numbers[]{1000001, 1000002, 1000003, 1000004, 1000005, 1000006, 1000007, 1000008, 1000009};
```
bernhardmgruber, line 3: Even though the `constexpr` is fine here, it confused some students during my afternoon session. So I propose to simplify it to a `const`.

code/control/control.cpp
```diff
@@ -1,12 +1,12 @@
...
 
 bool isodd(unsigned int i) { return i % 2 == 1; }
 
 void part1() {
-  int sum_odd = 0;
-  int sum_eve = 0;
+  unsigned int sum_odd = 0;
+  unsigned int sum_eve = 0;
```
bernhardmgruber, line 9: Otherwise CLion shows a warning here because I mix `int` and `unsigned`.

code/functions/README.md
```diff
@@ -22,4 +22,6 @@ Try passing by copy and passing by reference, see the difference.
 
 ## Step 4
 
-When passing by reference, ensure that your `printFiveCharacters` cannot inadvertently modify the original object. To test its const correctness, try adding something like `argument.data[0] = 'a';` to your print function. Try both with and without const attributes in your print function's signature.
+When passing by reference, ensure that your `printFiveCharacters` cannot inadvertently modify the original object.
+To test its const correctness, try adding something like `argument.name[0] = 'a';` to your print function.
+Try both with and without const attributes in your print function's signature.
```
bernhardmgruber, line 27: Main change is `argument.data[0]` turned to `argument.name[0]`.

sponce (approved):

PR #440, opened 2023-03-08
Simplify cmake setup

bernhardmgruber:
* Merge SolutionTarget.cmake and CompilerSettings.cmake into single file
* Require C++ standard


sponce (approved):

hageboeck (approved):
Thanks, good idea!

Issue #439, opened 2023-03-08
About operators exercise


chavid:
I found it too too long. Few suggestions to better focus on operators only:
- [x] Provide in the original code the normalize() function and std::gcd stuff.
- [ ] Provide int the original code the constructors, public attributes, constructors, `assign` method, and externals functions such as `multiply`, `add`, `compare` (returning -1,0,1 ?), `print`.
- [ ] Ask to replace the functions with operators: =, +, *, <=>, <<
- [ ] Then ask to add, reusing previous ones: !=, +=, *=.
- [ ] Show that with <=> implemented, then <, <=, >, >= are automatically infered ?
- [ ] Optionally, make all the assignments returning *this, like in operator<<, and explain this permit chaining.
- [ ] Optionally, make them move the operators as inline friend functions, and make attributes private.
- [ ] Optionnaly, discuss the pros and cons of making the unary constructor explicit.

bernhardmgruber:
I would like to add, that we should prefill the class declaration with two data members for numerator and denominator, probably as `int`. I had a few students implementing the class with a `float`.

Furthermore, I think we could remove `operator+` and `operator+=`, or move them to the back as bonus exercises.

bernhardmgruber:
My improvements are up here: #442

bernhardmgruber:
> obviously too hard for beginners who never wrote a single class.

Right. I think we should therefore have an exercise before where the students design their own (simple) class, like you say below.

>   * in the first one, we just learn to create a class with constructors, private attributes, accessors, and externals functions such as `multiply`, `add`, `compare`, `print`.

It does not need to be a `Fraction` class. It can be anything! We could also cover this as part of the polymorhism exercise, by asking the students to add a new class to the hierarchy.

> * provide in the original code the normalize() function and std::gcd stuff. thatś not the purpose to loose time with this.

Agreed, part of #442.

> * make them implement operators as external functions, relying on accessors, and do not discuss the option of implementing them as members, or friend functions. Beginners are not mature enough to understand this. better if they have time to play with tools and polymorphism, etc.

The exercise leaves it open how to implement the operators, by design. There is also no need to explain how hidden friends work. Students would only know if they looked at the solution, which they should not :)

> * do not talk about chaining =, +=. do it only for <<, it is largely hard enough for beginners. again, better if they have time to play with tools and polymorphism, etc. Or clearly advertise it as a final advanced optional step.

I removed the chaining of `operator*=` as well.

[...]

PR #442, opened 2023-03-08
Simplify operators exercise

bernhardmgruber:
* provide definition of normalize()
* provide two data members for numerator and denominator
* remove addition operators
* do not check whether *= can be chained

### DIFF
--- a/code/operators/operators.cpp
+++ b/code/operators/operators.cpp
@@ -1,18 +1,29 @@
 #include <iomanip>
 #include <iostream>
+#include <numeric>
 
 class Fraction {
+ public:
   // TODO: constructors and operators
+
+ private:
+  void normalize() {
+    const int gcd = std::gcd(m_num, m_denom);
+    m_num /= gcd;
+    m_denom /= gcd;
+  }
+
+  int m_num, m_denom;
 };
 
-// TODO:operators
+// TODO: operators
 
 
-void printAndCheck(const std::string & what, const Fraction & result, const Fraction & expected) {
+void printAndCheck(std::string const & what, Fraction const & result, Fraction const & expected) {
   const bool passed = result == expected;
   std::cout << std::left << std::setw(40) << what << ": " << (passed ? "PASS" : "** FAIL **") << "    " << result << "\n";
 }
-void printAndCheck(const std::string & what, bool result, bool expected) {
+void printAndCheck(std::string const & what, bool result, bool expected) {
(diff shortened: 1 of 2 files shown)


code/operators/operators.cpp
```diff
@@ -1,11 +1,23 @@
+#include <cassert>
```
hageboeck, line 1: Is this a remnant from a test?

code/operators/solution/operators.sol.cpp
```diff
@@ -72,11 +55,11 @@ class Fraction {
   int m_num, m_denom;
 };
 
-void printAndCheck(std::string const & what, Fraction const & result, Fraction const & expected) {
+void printAndCheck(const std::string & what, const Fraction & result, const Fraction & expected) {
```
sponce, line 58: I would really keep using const of the right of the type, as it's how it is suppose to work. For me const on the left should even be deprecated and ultimately forbidden

code/operators/solution/operators.sol.cpp
```diff
@@ -72,11 +55,11 @@ class Fraction {
...
   const bool passed = result == expected;
   std::cout << std::left << std::setw(40) << what << ": " << (passed ? "PASS" : "** FAIL **") << "    " << result << "\n";
 }
-void printAndCheck(std::string const & what, bool result, bool expected) {
+void printAndCheck(const std::string & what, bool result, bool expected) {
```
sponce, line 62: same here

sponce (approved):

hageboeck (commented):
Why is the addition removed? Too much of the same?

code/operators/solution/operators.sol.cpp
bernhardmgruber: Ah yes, this came back in because `operators.cpp` had west-const. I fill fix it.

If you want to have consistent formatting of the `const` keyword, maybe propose a `.clang-format` file. It can rearrange the const placement. I don't care so much, because I am really fine with west-const. It shows me `const` first, before the data type (or `auto`), which is more important to me.

bernhardmgruber:
> Why is the addition removed? Too much of the same?

Exactly. I think just implementing `operator*` already teaches enough about operators. `operator+` would just repeat that. On top, implementing `operator+(Fraction, Fraction)` is harder than the multiplication. And as @chavid pointed out in #439, should not be part of the exercise.

PR #443, opened 2023-03-09
Rename core modern C++ chapter to more language

bernhardmgruber:
Because the chapter is mostly about more language features. I would consider max. half of it "modern C++" in the sense of being C++11 and newer, so I think it does not justify the title.

### DIFF
--- a/talk/C++Course.tex
+++ b/talk/C++Course.tex
@@ -131,7 +131,7 @@
   \input{objectorientation/adl}
 \end{advanced}
 
-\section[More]{Core modern \cpp}
+\section[More]{More language}
 \input{morelanguage/constness}
 \begin{advanced}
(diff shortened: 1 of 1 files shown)


hageboeck:
I think renaming it is OK, but "more language" is very unspecific. I will think about an alternative name.

sponce:
:-D it was called more language before ! And the file still has the name actually. But it was felt too generic and changed to "core modern C++". I have no strong feeling, but that we should not spend our time looping on these 2 titles, which are also not so important

bernhardmgruber:
Alright, I interpret the closing of the ticket as: let's leave it like it is :)

sponce:
Yes, but if you really feel strong about it, just reopen. I do not really care and I just cleaned up after no update in 6 months

PR #444, opened 2023-03-09
Add a slide on the rule of zero

bernhardmgruber:
Since this was missed a few times when discussing the rule of 3/5, I concluded we need an extra slide for it.

### DIFF
--- a/talk/objectorientation/constructors.tex
+++ b/talk/objectorientation/constructors.tex
@@ -101,11 +101,9 @@
     };
   \end{cppcode}
   \pause
-  \begin{exampleblock}{The rule of 3/5/0 (\cpp98/\cpp11 and newer) - {\color{blue!50!white} \href{https://en.cppreference.com/w/cpp/language/rule_of_three}{cppreference}}}
-    \begin{itemize}
-    \item if a class has a destructor, a copy/move constructor or a copy/move assignment operator, it should have all three/five. strive for having none by using RAII types as members.
-    \end{itemize}
-  \end{exampleblock}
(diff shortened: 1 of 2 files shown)


talk/objectorientation/constructors.tex
```diff
@@ -101,11 +101,9 @@
...
   \end{cppcode}
   \pause
-  \begin{exampleblock}{The rule of 3/5/0 (\cpp98/\cpp11 and newer) - {\color{blue!50!white} \href{https://en.cppreference.com/w/cpp/language/rule_of_three}{cppreference}}}
-    \begin{itemize}
-    \item if a class has a destructor, a copy/move constructor or a copy/move assignment operator, it should have all three/five. strive for having none by using RAII types as members.
-    \end{itemize}
-  \end{exampleblock}
+  \begin{goodpracticeWithShortcut}{The rule of 3/5 (\cpp98/\cpp11) - {\color{blue!50!white}{Rule of 3/5} \href{https://en.cppreference.com/w/cpp/language/rule_of_three}{cppreference}}}
+    if a class needs a custom destructor, a copy/move constructor or a copy/move assignment operator, it should have all three/five.
+  \end{goodpracticeWithShortcut}
```
sponce, line 106: This breaks the good practice link (its title has become "I" and that I is missing from the content of the block. to be fixed.

Also I believe we will need to lower the font of the good practice slide, it starts to be too full, and 2 titles don't fit in a line (also to be fixed)

bernhardmgruber: Hmm, I think I just pasted something in the wrong place here. The `{Rule of 3/5}` should be the name of the shortcut.

PR #445, opened 2023-03-09
Allow for choosing essentials vs full via environment variable.

hageboeck:
Here is an attempt at switching between essentials and advanced course without changing files under version control. There is many ways now to switch to essentials:
- Uncomment a line in `C++Course.tex` (but git diff won't be empty)
- `export HEPCPP_ESSENTIALS=1`
  `make`
- `make essentials`
- `touch ./buildbasic`

For details, see the code proposed in this PR:
```latex
% There are three ways to switch to the basic course:
 % 1. Uncomment the line below. Disadvantage: git diff won't be empty
 %\basictrue
 % 2. Tell it to the Makefile.
 %  2.1 Invoke "make essentials"
 %  2.2 export HEPCPP_ESSENTIALS=1
 %      make <target>
 \ifdefined\makebasic
   \basictrue
 \fi
 % 3. Create a file named ./buildbasic
 \IfFileExists{./buildbasic}{\basictrue}
```

### DIFF
--- a/talk/setup.tex
+++ b/talk/setup.tex
@@ -92,6 +92,7 @@
 \usepackage{booktabs}
 \usepackage{comment}
 \usepackage{totcount}
+\usepackage{xspace}
 
 % Use C++Course.cut for output so it's cleaned by latexmk
 \def\DefaultCutFileName{\def\CommentCutFile{\jobname.cut}}
@@ -100,7 +101,7 @@
 %%%%%%%%%%%%%%%%%%%
 % useful commands %
 %%%%%%%%%%%%%%%%%%%
-\newcommand{\cpp}{C$^{++}$}
+\newcommand{\cpp}{C$^{++}$\xspace}
 \newcommand{\deprecated}{\textcolor{red}{\bf Deprecated}}
 \newcommand{\removed}{\textcolor{red}{\bf Removed}}
 
--- a/.github/workflows/build-slides.yml
+++ b/.github/workflows/build-slides.yml
@@ -22,11 +22,18 @@ jobs:
     steps:
       - name: Set up Git repository
         uses: actions/checkout@v3
-      - name: Set up essentials course
-        if: matrix.version == 'essentials'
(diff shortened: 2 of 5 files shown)


bernhardmgruber:
Hmm, this would break at least my workflow with TexStudio, which does not use the `Makefile` or any other command line tools. I just run the standard build commands via keyboard shortcuts from the IDE. I just checked, I could configure how latex is invoked from TexStudio, but that would also affect other projects, e.g. my dissertation or the paper I am writing as well from the same IDE.

I am also not a fan of the `\IfFileExists{onlybasics.tex}{\input{onlybasics}}{}`, but it has the nice benefit that I can choose whether to build the basic/advanced course without modifying any file under version control. I can see though that the environment variable works well for your workflow. Can we maybe keep a line with something like:
```latex
\IfFileExists{buildbasic}{\basictrue}{}
```
So the filecontent is irrelevant, but I have a second way to specify that I want the basic course? You can still change the CI of course!

hageboeck:
> Can we maybe keep a line with something like:
> `\IfFileExists{buildbasic}{\basictrue}{}`

That should work!

hageboeck:
> Can we maybe keep a line with something like:
> 
> ```
> \IfFileExists{buildbasic}{\basictrue}{}
> ```
> 
> So the filecontent is irrelevant, but I have a second way to specify that I want the basic course? You can still change the CI of course!

This should be done now, please check if it works for you.

sponce (approved):

talk/setup.tex
```diff
@@ -100,7 +101,7 @@
 %%%%%%%%%%%%%%%%%%%
 % useful commands %
 %%%%%%%%%%%%%%%%%%%
-\newcommand{\cpp}{C$^{++}$}
+\newcommand{\cpp}{C$^{++}$\xspace}
```
bernhardmgruber, line 104: Yeah, I played with `\xspace` as well, but I was too afraid to put it in the course slides yet :)

bernhardmgruber (approved):

PR #446, opened 2023-03-09
Mention more debug symbol flags

bernhardmgruber:
These flags keep popping up here and there when I watch C++ talks, so I wonder whether we should mention them as well. I have not tried them myself, since cmake usually chooses `-g` for me.

### DIFF
--- a/talk/tools/compiling.tex
+++ b/talk/tools/compiling.tex
@@ -70,7 +70,7 @@
   \end{block}
   \begin{block}{Optimization}
     \begin{description}
-      \item[-g] add debug symbols
+      \item[-g] add debug symbols (also: \texttt{-g3} or \texttt{-ggdb})
       \item[-Ox] 0 = no opt., 1-2 = opt., 3 = highly opt. (maybe larger binary), g = opt. for debugging
     \end{description}
   \end{block}


bernhardmgruber:
I also lerned that `-g` is just short for `-g2`, and there are also `-g0`, `-g1` etc. like the `-Ox` flags.

hageboeck:
I have never found a use for these options. If we mention them, we should also find out what we might need them for. Would you happen to have a reference?

bernhardmgruber:
> I have never found a use for these options. If we mention them, we should also find out what we might need them for. Would you happen to have a reference?

I saw it recently in this talk: https://www.youtube.com/watch?v=l3h7F9za_pc. Here is the gcc documentation: https://www.youtube.com/watch?v=l3h7F9za_pc. Specifically:

"-ggdb Produce debugging information for use by GDB. This means to use the most expressive format available (DWARF, stabs, or the native format if neither of those are supported), including GDB extensions if at all possible."
"-glevel Level 3 includes extra information, such as all the macro definitions present in the program. Some debuggers support macro expansion when you use -g3."

sponce:
I have to test -ggdb ! It's a long time I complain that gdb support is not as good as before. Maybe this may  explain !

PR #450, opened 2023-03-10
Small things during essentials day3

hageboeck:
Small items found during lecture.

### DIFF
--- a/talk/morelanguage/raii.tex
+++ b/talk/morelanguage/raii.tex
@@ -248,7 +248,7 @@ of 'std::unique_ptr<Foo>'
       void observe(const T&);
       void modifyRef(T&);
       void modifyPtr(T*);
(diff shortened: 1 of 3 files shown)


talk/morelanguage/raii.tex
```diff
@@ -253,7 +253,7 @@
       observe(*pt);                     // Keep ownership
       modifyRef(*pt);                   // Keep ownership
       modifyPtr(pt.get());              // Keep ownership
-      consume(std::move(pt));           // Transfer ownership
+      consumer(std::move(pt));          // Transfer ownership
```
bernhardmgruber, line 256: Since I renamed all the functions recently and forgot it here: The purpose was to use a verb rather than a noun as function name. So could you please rename the function to `consume` please? Thx!

bernhardmgruber (approved):

Issue #455, opened 2023-03-22
Add a script to check needed tools for hands-on exercises


bernhardmgruber:
It would be great to have a small script that could check whether the tools needed to run the hands-on exercises are installed and of a sufficiently up-to-date version. This would allow mentors, admins and also students to quickly check their setup.

bernhardmgruber:
Example output:
![image](https://user-images.githubusercontent.com/1224051/227516865-596d98d8-b34c-47d3-a6ed-79001179a900.png)

PR #457, opened 2023-03-24
Add a script to check your setup for the exercises

bernhardmgruber:
Fixes: #455

What's missing:

- [x] Check minimum version for some tools (compilers)
- [x] Handle alternative tools. Like you should either have `ldd`  or `otool`.

### DIFF
--- a/code/Makefile
+++ b/code/Makefile
@@ -1,7 +1,3 @@
-compile: *.tex
-	pdflatex -shell-escape $<
-	pdflatex -shell-escape $<
-
 TESTDIRS = callgrind cppcheck header_units control hello modules move python smartPointers templates virtual_inheritance \
 	   debug helgrind memcheck polymorphism race stl valgrind
 NOCOMPILETESTDIRS = constness
@@ -14,7 +10,6 @@ solution:
 	done
 
 clean:
-	rm -f *.aux *.log *.nav *.out *.pyg *.snm *.vrb *.toc
 	for dir in ${TESTDIRS} ${NOCOMPILETESTDIRS}; do \
 		cd $${dir}; \
 		make clean; \
@@ -22,4 +17,3 @@ clean:
 	done
 
 clobber: clean
-	rm -f *.pdf
--- a/code/check_setup.sh
+++ b/code/check_setup.sh
@@ -1,0 +1,61 @@
+#!/bin/bash
+
+RED='\033[0;31m'
+YEL='\033[0;33m'
+NORMAL='\033[0m'
+COL=12
+
+exit_status=0
+
+# from: https://stackoverflow.com/a/4024263
+function version_less() {
+  [ "$1" = "`echo -e "$1\n$2" | sort -V | head -n1`" ]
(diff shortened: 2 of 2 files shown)


bernhardmgruber:
On lxplus9:
image

hageboeck:
That's a great idea! Some more ideas:
- I would invert the meaning of the return codes, though. 0 signifies success, non-zero is a failure.
- The whole thing should also return success/failure
- I would use `II` to convey the logical "or" of tools. Here is my draft:
```bash
#!/bin/bash

RED='\033[0;31m'
YEL='\033[0;33m'
NORMAL='\033[0m'
COL=12

exit_status=0

# from: https://stackoverflow.com/a/4024263
function version_less() {
  [ "$1" = "`echo -e "$1\n$2" | sort -V | head -n1`" ]
}

function has_tool() {
  min_version=$2 || "0"
  required=$3 || true
  if [ -x "$(command -v $1)" ]
  then
    version=$($1 --version | grep -o -E "[0-9]+.[0-9]+(.[0-9]+)?" | head -1)
    if version_less $version $min_version
    then
      printf "%-${COL}s found, ${RED}version: ${version}, need at least: ${min_version}${NORMAL}\n" $1
    else
      printf "%-${COL}s found, version: ${version}\n" $1
      return 0
    fi
  elif $required
  then
    printf "${RED}%-${COL}s not found, but required${NORMAL}\n" $1
    exit_status=1
  else
    printf "${YEL}%-${COL}s not found, but optional${NORMAL}\n" $1
  fi
  return 1
}

# compiler
has_tool g++ 18.0.0 false || has_tool clang++ 18.0.0 false || { echo -e "${RED}No supported compiler found${NORMAL}"; exit_status=1; } # clang does not work for header_units exercise

# build tools
has_tool make
has_tool cmake 3.12.0
has_tool ccmake 3.12.0 false

# debugger
has_tool gdb 10.0.0 false || has_tool lldb 11.0.0 false || { echo -e "${RED}No supported debugger found${NORMAL}"; exit_status=1; }

# utils
has_tool nm
has_tool ldd 0.0 false || has_tool otool 0.0 false || { echo -e "${RED}Missing ldd or otool${NORMAL}"; exit_status=1; }

# tools
has_tool valgrind
has_tool kcachegrind
has_tool cppcheck
has_tool clang-format
has_tool clang-tidy
has_tool python3

exit $exit_status
```

bernhardmgruber:
> * I would invert the meaning of the return codes. 0 signifies success, non-zero is a failure.

That's what I had before, but then I had to negate in a lot of places and thought I should formulate the function as a query, hence the name `has_tool`, which should obviously return `true` on success. But you are right, this conflicts with the error code convention.

> * The whole thing should also return success/failure

Good idea!

> * I would use `||` to convey the logical "or" of tools. In this case, the function `has_one_of` should be obsolete.

Yeah, this is stuff that I don't know yet :)

hageboeck (approved):
LGTM, thanks

PR #464, opened 2023-05-15
Make function exercise print only 5 chars

bernhardmgruber:
This PR points out in the instructions of the functions exercise that the solution should only print 5 characters in the end. The solution is adapted with a few alternatives on how to solve this.


hageboeck (approved):
LGTM, but maybe we can squash on merge.

PR #469, opened 2023-05-16
Add some useful gdb commands

bernhardmgruber:
Students asked me to list the more important commands to solve the debug exercise. Here is what I collected.

How should we best present these? Here in the README.md of the exercise? And what should we do about `lldb`?

### DIFF
--- a/code/debug/README.md
+++ b/code/debug/README.md
@@ -5,3 +5,24 @@
 * inspect backtrace, variables to understand the problem
 * try stepping through the code using 'step' and 'next', use breakpoints and continue to get familiar with gdb
 * fix the bug
+
+### Some useful gdb commands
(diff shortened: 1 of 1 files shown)


bernhardmgruber:
Mention this helper for migrating from gdb to lldb: https://lldb.llvm.org/use/map.html. Courtesy of @pcanal.

hageboeck:
Very useful commands, indeed. I would add
```
info br: list active breakpoints
enable/disable/delete #: Enable/disable/delete breakpoints
```

bernhardmgruber:
> Very useful commands, indeed. I would add

Done, thx!

hageboeck (approved):
Good to merge for me.

PR #471, opened 2023-05-16
Simplify debug exercise

hageboeck:
In addition to the work in #468, I propose to simplify the code. It seemed to be complicated enough that we didn't fix an out-of-bounds access in the solution, so here I propose to take the randomness out of the exercise, and go for a simple reverse.

During the C++ course at JLab, a mentor suggested to add a check for correctness, which is trivial when the collection just has to be reversed.

I did a similar thing for the valgrind part, but I added an off-by-one mistake in the reverse function, and a memory leak in main. That should hopefully be enough for people to find.

Note: Either close #468 and merge the current one, or merge #468 and this gets rebased.


code/debug/debug.cpp
```diff
@@ -0,0 +1,39 @@
...
+void swap(int *a, int*b) {
+    int c = *a;
+    *a = *b;
+    *b = c;
+}
+
+void reverse(int *v, unsigned int len) {
+    for (unsigned int i = 0; i < (len + 1) / 2; i++) {
+      const int a = i;
+      const int b = len - 1 - i;
```
bernhardmgruber, line 13: So my wristband convinced you today? :)

bernhardmgruber (approved):

bernhardmgruber (changes requested):
Please update the `README.md` of the debug exercise as well!

sponce (approved):

bernhardmgruber (approved):

code/debug/debug.cpp
```diff
@@ -0,0 +1,39 @@
...
+      v[i] = i;
+    }
+}
+
+int main() {
+    constexpr auto arraySize = 100;
+    int *v = nullptr;
+    // create and reverse the vector of LEN numbers
+    reverse(v, 1000);
+    createAndFillVector(v, arraySize);
```
pcanal, line 31:
```suggestion
    v = createAndFillVector(arraySize);
```

code/debug/debug.cpp
```diff
@@ -0,0 +1,39 @@
...
+      swap(v + a, v + b);
+    }
+}
+
+void createAndFillVector(int*& v, unsigned int len) {
+    v = new int[len];
+    for (unsigned int i = 0; i < len; i++) {
+      v[i] = i;
+    }
+}
```
pcanal, line 24:
```suggestion
int *createAndFillVector(unsigned int len) {
    auto v = new int[len];
    for (unsigned int i = 0; i < len; i++) {
      v[i] = i;
    }
    return v;
}
```

code/debug/debug.cpp
hageboeck: Very good, thanks! It goes in the direction of value semantics, so I will do this transformation also with the other three files.

PR #475, opened 2023-05-18
Improve clarity of constness exercise

bernhardmgruber:
* Rename functions to more meaningful names
* Extent instructions to tell students to only change lines inside the main function

### DIFF
--- a/code/constness/constplay.cpp
+++ b/code/constness/constplay.cpp
@@ -1,6 +1,10 @@
 #include <iostream>
 #include <string>
 
+/* This is a dummy function to demonstrate pass by value.
+ * Since it doesn't do anything with the argument, we suppress
+ * possible compiler warnings using `maybe_unused`.
+ */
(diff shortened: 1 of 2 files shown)


code/constness/constplay.cpp
```diff
@@ -1,23 +1,23 @@
 #include <iostream>
 #include <string>
 
-int identity(int a) {
-    return a;
+void copy(int a) {
+    [[maybe_unused]] int val = a;
```
hageboeck, line 5:
```suggestion
/* This is a dummy function to demonstrate pass by value.
 * Since it doesn't do anything with the argument, we suppress
 * possible compiler warnings using `maybe_unused`.
 */
void copy(int a) {
    [[maybe_unused]] int val = a;
```

hageboeck (changes requested):
I tried to clarify things a bit more.

hageboeck (approved):

PR #476, opened 2023-05-18
Clarify instructions of template exercise

bernhardmgruber:


code/templates/README.md
```diff
@@ -4,7 +4,7 @@
 Beginners
 * Look at the `OrderedVector` code
 * Compile and run `playwithsort.cpp`. See the ordering
-* Modify `playwithsort.cpp` and reuse `OrderedVector` with `Complex`
+* Extend `playwithsort.cpp` and demonstrate that `OrderedVector` works with `Complex` as element type as well.
```
hageboeck, line 7:
```suggestion
* In `playwithsort.cpp`, create a second `OrderedVector` with `Complex` as element type. The class should compile without modifications.
```

code/ExercisesCheatSheet.md
```diff
@@ -109,9 +109,10 @@ See in valgrind the improvements.
 
 This exercise has several levels. People not at ease can stop after first level and go to next exercise. Alternatively, they may do level 1 and 3 and skip 2.
 
-Level 1 : just use the given Complex class in OrderedVector and see it works out of the box thanks to generic code in OrderedVector.
+Level 1: just use the given `Complex` class in `OrderedVector` and see it works out of the box thanks to generic code in `OrderedVector`.
+Make sure, that students don't create a `Complex<std::string>`.
```
hageboeck, line 113: Please explain why they might do that. I don't understand what I'm supposed to do as mentor. Imagine this:
"The next one is the templates exercise. Don't use `Complex<std::string>`, though!"

I think the students will be as confused as I am now.

bernhardmgruber: I have literally no idea. Around half of my students in the JLab lecture started to create `Complex<std::string>` and were struggling with compilation errors. I don't know where they picked up this idea. I am trying to massage the instructions to avoid future students to fall into this pit.

bernhardmgruber: I removed it again and massaged the other instrucitons a little bit and added a requires clause to `Complex<T>`.

code/templates/Complex.hpp
```diff
@@ -1,6 +1,8 @@
 #include <ostream>
+#include <concepts>
 
 template <typename T=float>
+    requires std::floating_point<T>
```
bernhardmgruber, line 5: I intentionally did not write it as:
```c++
template <std::floating_point T=float>
```
So it's less unknown syntax for the students.

bernhardmgruber:
Hmm, do we really want to show:
```c++
    std::cout << "Int Complex\n";
    using IComplex = Complex_t<int>;
    OrderedVector<IComplex> vci(5);
...
    std::cout << "Very Complex\n";
    using VComplex = Complex_t<Complex>;
    OrderedVector<VComplex> vcv(5);
```
in the solution?

code/ExercisesCheatSheet.md
hageboeck: Maybe instead of saying:
Don't do this

we could say:
To do this, start by xxxx

code/templates/Complex.hpp
hageboeck: Do we support concepts? This might not compile for some students.

A way out would be a feature test, but that might confuse them even more? 🤔

bernhardmgruber: The exercise already compiled with C++20 before, so I thought it was fine.

But this change breaks the solution, which also does `OrderedVector<Complex_t<Complex>>`, and `Complex` is not a floating point. Hence the question whether we actually want to support this below.

bernhardmgruber: I removed it again now. The `Comple_t<Complex>` is a bonus exercise, so the `requires` clause would ruin it.

code/ExercisesCheatSheet.md
bernhardmgruber: I extended the instructions and cheatsheet a bit.

bernhardmgruber:
> Hmm, do we really want to show:
> 
> ```c++
>     std::cout << "Int Complex\n";
>     using IComplex = Complex_t<int>;
>     OrderedVector<IComplex> vci(5);
> ...
>     std::cout << "Very Complex\n";
>     using VComplex = Complex_t<Complex>;
>     OrderedVector<VComplex> vcv(5);
> ```
> 
> in the solution?

Yes, it's the bonus exercise ..

hageboeck:
> Hmm, do we really want to show:
> 
> ```c++
>     std::cout << "Int Complex\n";
>     using IComplex = Complex_t<int>;
>     OrderedVector<IComplex> vci(5);
> ...
>     std::cout << "Very Complex\n";
>     using VComplex = Complex_t<Complex>;
>     OrderedVector<VComplex> vcv(5);
> ```
> 
> in the solution?

I guess with explanation it's OK. Mathematically, it might not make sense, but you can sort them ...

hageboeck (approved):
LGTM

PR #477, opened 2023-05-18
Add a clang-format style file.

hageboeck:
Here is a draft of a clang-format style based on WebKit.

I propose to agree on a style and merge the `.clang-format` file without changing every single file immediately.
People editing with clang-format-capable editors will automatically check in formatted code when they work on a file.
After a while, when we are happy with the style it imposes, we can write a loop over all files and create a commit.

Four files have already been formatted using the draft style, see:
- [`debug/debug.cpp`](https://github.com/hsf-training/cpluspluscourse/blob/e5de9717446dc01671579357ac8f3b7802d09bed/code/debug/debug.cpp) and solution
- or `valgrind/debug.cpp` and solution

### DIFF
--- a/code/.clang-format
+++ b/code/.clang-format
@@ -1,0 +1,3 @@
+BasedOnStyle: WebKit
+InsertNewlineAtEOF: True
+Standard: Latest
--- a/code/debug/debug.cpp
+++ b/code/debug/debug.cpp
@@ -1,44 +1,44 @@
 #include <algorithm>
 #include <iostream>
 
-void swap(int * a, int * b)
+void swap(int* a, int* b)
 {
-  int c = *a;
-  *a = *b;
-  *b = c;
+    int c = *a;
+    *a = *b;
+    *b = c;
 }
 
-void reverse(int * v, unsigned int len)
+void reverse(int* v, unsigned int len)
 {
-  for (unsigned int i = 0; i < (len + 1) / 2; i++) {
-    const int a = i;
-    const int b = len - 1 - i;
+    for (unsigned int i = 0; i < (len + 1) / 2; i++) {
+        const int a = i;
+        const int b = len - 1 - i;
 
-    swap(v + a, v + b);
-  }
+        swap(v + a, v + b);
+    }
 }
 
-int * createAndFillVector(unsigned int len)
+int* createAndFillVector(unsigned int len)
 {
-  auto v = new int[len];
-  for (unsigned int i = 0; i < len; i++) {
-    v[i] = i;
-  }
-  return v;
+    auto v = new int[len];
+    for (unsigned int i = 0; i < len; i++) {
+        v[i] = i;
+    }
+    return v;
 }
 
 int main()
 {
-  constexpr auto arraySize = 100;
-  int * v = nullptr;
-  // create and reverse the vector of LEN numbers
-  reverse(v, 1000);
-  v = createAndFillVector(arraySize);
+    constexpr auto arraySize = 100;
+    int* v = nullptr;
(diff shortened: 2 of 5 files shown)


sponce:
In my experience (in both acceptations of the terms actually), this is more tricky than it seems, as the indentation will vary with versions of clang... Thus if you have locally a different version than I do, we'll fight about indentation. The consequence is that the indentation can only be enforced in the gitlab CI, which means that it should fail when indentation is not fine and give a URL where to download the diff that people will apply by hand locally before they force push... Really heavy, but the way we work in LHCb

Another possibility is of course to share a common version of clang, but then we have to take it e.g. from cvmfs which means that you lose the ability to work offline.

As I said, no so easy... Any strong feeling in which direction we should go ?

bernhardmgruber:
> the indentation can only be enforced in the gitlab CI, which means that it should fail when indentation is not fine and give a URL where to download the diff that people will apply by hand locally before they force push...

Wow, this is exactly how [LLAMA CI](https://github.com/alpaka-group/llama/blob/develop/.github/workflows/ci.yaml#L18-L36) works :D :D I also implemented it the same way for alpaka and a few other projects. What does not work great is to push the clang-format changes to the branch from the CI. That fails sometimes.

> Another possibility is of course to share a common version of clang, but then we have to take it e.g. from cvmfs which means that you lose the ability to work offline.

Hmm. I get my clang from LLVM's apt, so I have the newest version a day after it was released :) I have seen active projects at CERN still on clang-format 3.x something, because they agreed on a common version. So maybe let's not go there :)

> As I said, no so easy... Any strong feeling in which direction we should go ?

For now I would do like @hageboeck said. Don't enforce it, but have a style file for new code or when we modify exercises. We can then fix the style at some point in the future and add a CI run. It's not too hard. And clang-format is mostly stable across versions now. I see like 1 of a few 1000 lines change whenever I upgrade. I think we can tolerate that.

code/.clang-format
```diff
@@ -0,0 +1,6 @@
+BasedOnStyle: WebKit
+IndentWidth: 2
```
bernhardmgruber, line 2: I prefer 4 here, but I would rather have a any `.clang-format` file than none. So go ahead. Just if someone else thinks my way.

Do you think 2 is better for presenting exercises on projectors for group discussions? May actually be a good idea.

code/.clang-format
```diff
@@ -0,0 +1,6 @@
...
+InsertNewlineAtEOF: True
+PointerAlignment: Middle
```
bernhardmgruber, line 4: I personally prefer left here, but right is also common in some code bases. So yeah, maybe middle is the right way to mess with both types of people :)

sponce:
> And clang-format is mostly stable across versions now. I see like 1 of a few 1000 lines change whenever I upgrade. I think we can tolerate that.

You indeed have a point here. I see the same, which is clearly a drama with a project containing > 10M lines of code. but of course we are not in that case, so here it should not be too bad indeed. Let's try.

bernhardmgruber (approved):
I have not run the file for now, but it's good to have one. I can suggest modifications in later PRs if I would be really unhappy.

code/.clang-format
sponce: I would just stick the the default (also for the number of spaces). We can have the usual war on tastes, but the easiest is clearly to have no change at all to the default, so drop all these extra lines after WebKit.

hageboeck: OK for me to drop all the lines, but we need the new line at the end of the file. Otherwise, the pre-commit CI will add a commit to every PR where this is not the case. I will apply the changes and merge afterwards.

Issue #456, opened 2023-03-24
Rename the `code` directory to `exercises`


bernhardmgruber:
I always have a 4 second brain hang when I try to find the exercises folder because `code` just sounds too generic. I propose to rename the folder to `exercises`. Any objections?

bernhardmgruber:
Just happended again. @sponce and @hageboeck, what do you think?

sponce:
I do not care. Just remember you have to change ALL places where this is referred to. Please git grep

PR #478, opened 2023-06-21
Rename directory `code` to `exercises`

bernhardmgruber:
Fixes: #456


hageboeck (approved):
Big change. 🙂 

I cannot possibly check if you missed something on gh, so I proposed to merge, and you will forever be in our debt if you overlooked one. 😉

bernhardmgruber:
> Big change. slightly_smiling_face
> 
> I cannot possibly check if you missed something on gh, so I propose to merge, and you will forever be in our debt if you overlooked one. wink

I will carry this responsibility with honour!

PR #479, opened 2023-06-23
Discuss "using namespace std;"

### LINKED ISSUE #461: We don't discuss `using namespace std`
We got a question about why you would use `using namespace std`, and not explained in the slides.

What would be a good place to discuss it? Just at the end of the section about namespaces? We didn't really discuss include and standard library at this point yet ...

hageboeck:
Add a slide on using namespace std and when not to use it. The slide is at the end of the namespace chapter in the advanced course.

I'm not very sure about the placement of the slide. We have discussed scopes and namespaces, but almost no `std` stuff yet. Also header/implementation haven't been discussed. I tried not to use a lot of `std` knowledge, but it's unavoidable.
Can you think of a location later in the course where this slide would make sense?

Fix #461

### DIFF
--- a/talk/basicconcepts/scopesnamespaces.tex
+++ b/talk/basicconcepts/scopesnamespaces.tex
@@ -183,4 +183,33 @@ int a = 1;
     \end{cppcode*}
   \end{alertblock}
 \end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[98]{Using namespace directives}
+  \begin{alertblock}{Avoid ``using namespace'' directives}
+    \begin{itemize}
+      \item Make all members of a namespace visible in current scope
+      \item Risk of name clashes or ambiguities
+    \end{itemize}
+    \begin{cppcode*}{}
+      using namespace std;
+      cout << "We can print now\n"; // uses std::cout
+    \end{cppcode*}
+  \end{alertblock}
+  \begin{alertblock}{Never use in headers at global scope!}
+    \begin{cppcode*}{gobble=2}
+      #include "PoorlyWritten.h" // using namespace std;
+      struct array { ... };
+      array a;  // Error: name clash with std::array
+    \end{cppcode*}
+  \end{alertblock}
+  \begin{block}{What to do instead}
+    \begin{itemize}
+      \item Qualify names: \cppinline{std::vector}, \cppinline{std::cout}, \ldots
+      \item Put things that belong together in the same namespace
+      \item Use \textit{using declarations} in local scopes: \cppinline{using std::cout;}
+    \end{itemize}
+  \end{block}
+\end{frame}
+
 \end{advanced}


talk/basicconcepts/scopesnamespaces.tex
```diff
@@ -183,4 +183,31 @@
...
 \end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[98]{Using namespace std}
+  \begin{block}{Purpose}
+    \begin{itemize}
+      \item Directive to make all members of the \texttt{std} namespace visible
+      \item Can be used for convenience, but risk of name clashes
+      \item OK in implementation files or scopes
+      \item Alternative: using declaration (e.g.\ \cppinline{using std::cout;})
```
bernhardmgruber, line 194:
```suggestion
      \item Directive to make all members of the \texttt{std} namespace visible
      \item Can be used for convenience, but risks name clashes
      \item OK in implementation files or inside scopes
      \item Alternative: using declaration (e.g.\ \cppinline{using std::cout;})
```

bernhardmgruber, line 194: Would you say such a using declaration is fine in a public header? I don't think so, but it is for sure less dangerous.

I am afraid that the slide could be interpreted as: Use using declarations instead in a header file!

talk/basicconcepts/scopesnamespaces.tex
```diff
@@ -183,4 +183,31 @@
...
+    \end{cppcode*}
+  \end{exampleblock}
+  \begin{alertblock}{Never use in a public header!}
+    \begin{cppcode*}{gobble=2,firstnumber=4}
+      #include "EvilHeader.h" // has using namespace std;
+
+      struct array { ... };
+      array a;  // Error: name clash with std::array
+    \end{cppcode*}
+  \end{alertblock}
```
bernhardmgruber, line 210:
```suggestion
  \begin{alertblock}{Never use in a header at global scope!}
    \begin{cppcode*}{gobble=2,firstnumber=4}
      #include "EvilHeader.h" // has using namespace std;

      struct array { ... };
      array a;  // Error: name clash with std::array
    \end{cppcode*}
  \end{alertblock}
```

bernhardmgruber:
> I'm not very sure about the placement of the slide. We have discussed scopes and namespaces, but almost no `std` stuff yet. Also header/implementation haven't been discussed. I tried not to use a lot of `std` knowledge, but it's unavoidable. Can you think of a location later in the course where this slide would make sense?

Since the rule applies more generally to using other namespaces in headers as well (`std` is just the prime example), I think it's fine that we did not discuss the standard library in detail yet.

The chapter on scopes and namespaces is the only place where we discuss `using namespace ...`, so I think it is a good place to discuss `using namespace std;` at the end!

talk/basicconcepts/scopesnamespaces.tex
hageboeck: That's not what I meant, indeed. Needs rephrasing.

sponce:
Honestly I would not mention this at all, or if I would it would be to strongly recommend to NEVER use `using namespace std`. To me it's certainly not "OK in implementation files or inside scopes". You get the same clashes and vicious bugs as anywhere else.

As a correlate we should also strongly recommend to not use at all `using namespace`. Such a need is a clear indication of a flow in the hierarchy of namespaces put in place. If you're in the right namespace, you won't need it.

bernhardmgruber:
The thing is, people at the JLab course asked about this specifically during the morning lecture and we discussed it. So I think we should have a slide about.

> As a correlate we should also strongly recommend to not use at all using namespace. Such a need is a clear indication of a flow in the hierarchy of namespaces put in place. If you're in the right namespace, you won't need it.

I agree with this philosophy and it's usually also my practice. So maybe we should drop the `\item OK in implementation files or scopes`.

hageboeck:
> Honestly I would not mention this at all, or if I would it would be to strongly recommend to NEVER use `using namespace std`. To me it's certainly not "OK in implementation files or inside scopes". You get the same clashes and vicious bugs as anywhere else.
> 
> As a correlate we should also strongly recommend to not use at all `using namespace`. Such a need is a clear indication of a flow in the hierarchy of namespaces put in place. If you're in the right namespace, you won't need it.

I would say that there is an exception to "you don't need it":
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c165-use-using-for-customization-points

I agree that ADL + using is a very advanced topic, so we should go simpler here and first explain what it does and then that it should be used only very sparingly ...
Is it OK if I remove the statement that "It is OK", and instead we warn to not overdo it?

hageboeck:
@sponce @bernhardmgruber 
I think now it's cautioning people to not overly used it. This slide hopefully gives enough material to allow for a good explanation during the lecture.
OK to merge?

talk/basicconcepts/scopesnamespaces.tex
hageboeck: Rephrasing done.

sponce:
Kind of ok for me now. Although I would name it "using namespace directive" and be slightly more strict, saying : "Avoid as much as you can" and adding "Prefer raw `using`"

Note that I disagree with your point stating "there is an exception". I really believe `using namespace` should never be used. Now `using` can of course, no problem there for a targeted import you know won't clash. You'll notice by the way that in your pointer https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c165-use-using-for-customization-points that's what they use, not `using namespace`

hageboeck:
> Note that I disagree with your point stating "there is an exception". I really believe `using namespace` should never be used. Now `using` can of course, no problem there for a targeted import you know won't clash. You'll notice by the way that in your pointer https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c165-use-using-for-customization-points that's what they use, not `using namespace`

We don't disagree. I had just misunderstood your comment. 🙂
Since we all don't use using directives (`using namespace`), but use using *declarations*, I'll do one final rephrasing. The slide title is already changed as you suggested.

sponce:
I must say that I'm a bit lost in the vocabulary here : `directive` means namespace and `declaration` means no namespace, that's it ? Then maybe make it clear, e.g. by adding to the title '(aka using namspace)' for example.

bernhardmgruber:
> I must say that I'm a bit lost in the vocabulary here : `directive` means namespace and `declaration` means no namespace, that's it ?

Exactly. See also: https://en.cppreference.com/w/cpp/keyword/using

sponce:
I like the last version ! Maybe just a detail if it fits on the slide, I would split into 2 lines the advice "Qualify names (\cppinline{std::vector}), put things that belong together in the same namespace", that is :
  - Qualify names (\cppinline{std::vector})
  - put things that belong together in the same namespace

sponce (approved):

bernhardmgruber (approved):

PR #480, opened 2023-06-28
Fix a tikz positioning directive that points to a non-existing element.

hageboeck:
I was playing with tikz, and this showed up as error. As far as I can see, the "title" object was removed from that figure.


bernhardmgruber (approved):
I assume you checked whether the slide looks the same :)

hageboeck:
> I assume you checked whether the slide looks the same :)

Actually, no, but I checked that it looks correct. 😄 

image

PR #488, opened 2023-08-30
Dropped useless bigdata member in the functions example

sponce:
It served no purpose (copy was not slow there, only a sleep made it slow) and was discovered (the hard way) to make windows users having very strange segfaults. Simply because it was too large to be on the stack in such context and was screwing memory !


bernhardmgruber:
Haha, AFAIK Visual C++ on Windows uses a default stack size of 1MiB. I am surprised we never ran into this.

However, now it's less obvious that `SlowToCopy` is slow because it's "big". Would it work if you just reduce the size of the data member? Maybe just to `int bigdata[10'000];`.

bernhardmgruber (approved):

PR #493, opened 2023-10-11
Tiny fixes found while preparing next session of the advanced course

sponce:


talk/basicconcepts/assert.tex
```diff
@@ -77,7 +77,7 @@
     \begin{cppcode*}{}
       double f(UserType a) {
         static_assert(
-          std::is_floating_point<UserType>::value,
+          std::is_floating_point_v<UserType>,
```
bernhardmgruber, line 80: Requires C++17 now, but the slide says 11. I guess this was the reason, the former version was used. But I totally agree that we should teach the students to use the `_v` traits!

bernhardmgruber (approved):
Small nit, otherwise good!

talk/basicconcepts/assert.tex
sponce: Well spotted ! I added a comment "C++17" on the given line, leaving 11 at the top as it targets the static_assert there.

bernhardmgruber: Parfait!

bernhardmgruber (approved):

PR #494, opened 2023-10-11
Added on slide on std::expected

sponce:


talk/morelanguage/morestl.tex
```diff
@@ -147,6 +147,35 @@
...
+      \end{itemize}
+    \end{itemize}
+  \end{block}
+  \begin{exampleblock}{}
+    \small
+    \begin{cppcode*}{}
+      enum class Error {...};
+      std::expected<int, Error> parse(std::string_view in) {
+        if (is_valid(in)) return convert_to_int(in);
+        return std::expected(Error::...);
```
bernhardmgruber, line 167:
```suggestion
        return std::unexpected(Error::...);
```
For the un-happy path, I think the intention is to use `std::unexpected`, although I have no experience with `std::expected` yet :)

This is also shown in the example on cppreference: https://en.cppreference.com/w/cpp/utility/expected

PR #498, opened 2023-10-16
Small fixes for advanced course

hageboeck:
Here are the fixes we discovered during the autumn 2023 lectures of the advanced course.


sponce:
For the last change (ADL stuff), I also did it in #499 and I prefer my version (looks more precise to me). Can you review it and if you're happy with it, drop yours ? Alternatively modify mine with yours of course

PR #501, opened 2023-10-18
fixes from 2nd day of Oct 23 session

sponce:
includes a move of the `if constexpr` slide from STL to constexpr subchapter


hageboeck (approved):
LGTM!

PR #508, opened 2024-01-19
Implement comments from #428

### LINKED ISSUE #428: Items from 6th and JLab C++ Course
- [ ]  ~~Add a slide on integral promotion~~

hageboeck:
This is the implementation of many remarks / suggestions made in #428, such as:
- Better explain header/source separation
- Add some code highlights or drawings
- ~~Discuss object slicing~~
- Add possible operator symbols for overloading
- Move `this` slide to a later stage of the OO chapter (and change the confusing code example)

I decided to disregard many comments from #428, because there was no reasonable way to change the slides. These comments referred rather to what we say or questions that the students asked. All the things I did implement are marked with tick boxes, and the things I left out have empty boxes in #428. I don't see a way to do better, so I propose to close the issue after this PR.

Resolves #428

### DIFF
--- a/talk/basicconcepts/headersinterfaces.tex
+++ b/talk/basicconcepts/headersinterfaces.tex
@@ -73,3 +73,23 @@
     \end{cppcode*}
   \end{block}
 \end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[98]{Header / source separation}
+  \begin{goodpractice}[Header / source]{Headers and source files}
+    \begin{itemize}
+      \item Headers should contain declarations of functions / classes
+        \begin{itemize}
+          \item Only create them if interface is used somewhere else
+        \end{itemize}
+      \item Might be included/compiled many times
+      \item Good to keep them short
+      \item Minimise \cppinline{#include} statements
+      \item Put long code in implementation files. Exceptions:
+        \begin{itemize}
+          \item Short functions
+          \item Templates and constexpr functions
+        \end{itemize}
+    \end{itemize}
+  \end{goodpractice}
+\end{frame}
--- a/talk/objectorientation/advancedoo.tex
+++ b/talk/objectorientation/advancedoo.tex
@@ -1,6 +1,45 @@
 \subsection[advOO]{Advanced OO}
 
 \begin{frame}[fragile]
+  \frametitlecpp[98]{\texttt{this} keyword}
+  \begin{block}{How to know an object's address?}
+    \begin{itemize}
+    \item Sometimes we need the address of the current object
+    \item Or we need to pass our address / a reference to a different entity
+      (for example to implement operators, see later)
+    \item All class methods can use the keyword \cppinline{this}
+      \begin{itemize}
+        \item It returns the address of the current object
+        \item Its type is \cppinline{T*} in the methods of a class {\ttfamily T}
+      \end{itemize}
+    \end{itemize}
+  \end{block}
+  \begin{minipage}{0.7\textwidth}
+  \begin{cppcode}
+    struct S {
+      int a,b;
+      // these two are the same:
+      int getB() { return b; }       // 5
+      int getB() { return this->b; } // 5
+      void testAddress() {
+        S* addr = this; // 0x3000
+      }
+    } s{2,5};
+  \end{cppcode}
+  \end{minipage}%
+  \hfil%
+  \begin{minipage}{0.3\textwidth}
+      \begin{tikzpicture}
+        \memorystack[size x=2cm,word size=1,block size=4,nb blocks=4]
+        \memorypush{a = 2}
+        \memorypush{b = 5}
+        \memorystruct{1}{2}{\tiny s}
+        \draw[-Triangle,thick] (\stacksizex-1*\stacksizey,-1*\stacksizey) node[left] {\footnotesize \cppinline{this} pointer} -- (\stacksizex,-0.1*\stacksizey);
+      \end{tikzpicture}
+    \end{minipage}
+\end{frame}
+
+\begin{frame}[fragile]
   \frametitlecpp[98]{Polymorphism}
   \begin{block}{the concept}
     \begin{itemize}
--- a/talk/objectorientation/advancedoo.tex
+++ b/talk/objectorientation/advancedoo.tex
@@ -571,15 +571,18 @@
   \frametitlecpp[98]{Multiple inheritance advice}
   \begin{goodpractice}{Avoid multiple inheritance}
     \begin{itemize}
-    \item Except for inheriting from interfaces
-    \item And for rare special cases
+      \item Except for inheriting from interfaces (=no data members)
+      \item And for rare special cases
     \end{itemize}
-  \end{goodpractice}
-  \begin{goodpractice}[NO diamond inheritance]{Absolutely avoid diamond-shaped inheritance}
-    \begin{itemize}
-    \item This is a sign that your architecture is not correct
-    \item In case you are tempted, think twice and change your mind
-    \end{itemize}
+
+    \hspace*{0.05\textwidth}\begin{minipage}{0.9\textwidth}
+      \begin{alertblock}{Absolutely avoid diamond-shape inheritance}
+        \begin{itemize}
+        \item This is a sign that your architecture is not correct
+        \item In case you are tempted, think twice and change your mind
+        \end{itemize}
+      \end{alertblock}
+    \end{minipage}
   \end{goodpractice}
 \end{frame}
 
--- a/talk/objectorientation/objectsclasses.tex
+++ b/talk/objectorientation/objectsclasses.tex
@@ -117,40 +117,18 @@
     \item some functions (templates, \cppinline{constexpr}) must be in the header
     \end{itemize}
   \end{goodpractice}
-  \begin{cppcode}
+  \begin{block}{}
+    \begin{cppcode}
+    #include "MyFirstClass.hpp"
+
     void MyFirstClass::squareA() {
       a *= a;
     }
-
     int MyFirstClass::sum(int b) {
       return a + b;
     }
-  \end{cppcode}
-\end{frame}
-
-\begin{frame}[fragile]
-  \frametitlecpp[98]{\texttt{this} keyword}
-  \begin{block}{How to know an object's address?}
-    \begin{itemize}
-    \item Sometimes we need to pass a reference to ourself to a different entity
-    \item For example to implement operators, see later
-    \item All class methods can use the keyword \cppinline{this}
-      \begin{itemize}
-        \item It returns the address of the current object
-        \item Its type is \cppinline{T*} in the methods of a class {\ttfamily T}
-      \end{itemize}
-    \end{itemize}
+    \end{cppcode}
   \end{block}
-  \begin{cppcode}
-    void freeFunc(S & s);
-    struct S {
-      void memberFunc() { // Implicit S* parameter
-        freeFunc(*this);  // Pass a reference to ourself
-      }
-    };
-    S s;
-    s.memberFunc();       // Passes &s implicitly
-  \end{cppcode}
 \end{frame}
 
 \begin{frame}[fragile]
--- a/talk/objectorientation/operators.tex
+++ b/talk/objectorientation/operators.tex
@@ -17,28 +17,36 @@
   \end{cppcode}
 \end{frame}
 
-\begin{frame}
+\begin{frame}[fragile]
   \frametitlecpp[98]{Operator overloading}
   \begin{block}{Defining operators for a class}
     \begin{itemize}
     \item implemented as a regular method
       \begin{itemize}
-      \item either inside the class, as a member function
+      \item \small either inside the class, as a member function
       \item or outside the class (not all)
       \end{itemize}
-    \item with a special name (replace @ by anything)
-      \begin{tabular}{llll}
-        Expression & As member & As non-member \\
-        \hline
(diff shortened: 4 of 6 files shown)


talk/basicconcepts/headersinterfaces.tex
```diff
@@ -73,3 +73,51 @@
...
+      \item Headers should only contain declarations that are necessary to know the interface of functions / classes
+      \item Headers might be included and compiled many times, so it is good to keep them short
+      \begin{itemize}
+        \item The same goes for \cppinline{#include} statements in headers
+      \end{itemize}
+      \item Implementation files only get compiled once, so put necessary includes and most of the code here
+        \begin{itemize}
+          \item Exception: Short functions can be in headers
+          \item Templates (see later) should be in headers
+        \end{itemize}
```
bernhardmgruber, line 90: Implementations of `constexpr` functions must be in headers. Maybe add that.

Also: every function where inlining would be a benefit, or where the function could be auto vectorized, is a good candidate for being in a header. But that is a big discussion with trade-offs and personal preferences, so hard to give a guideline here.

talk/basicconcepts/headersinterfaces.tex
```diff
@@ -73,3 +73,51 @@
...
+    \end{itemize}
+  \end{goodpractice}
+  \begin{block}{Header: Feature.hpp}
+    \begin{cppcode*}{}
+    #include "TypeDeclarations.hpp"
+    void feature(Type1 first, Type2 second);
+    \end{cppcode*}
```
bernhardmgruber, line 97: Random thought: if you pass by const reference, a forward declaration would suffice:
```latex
    \begin{cppcode*}{}
    class Type1;
    class Type2;
    void feature(const Type1& first, const Type2& second);
    \end{cppcode*}
```
This is sometimes forgotton, but may also not be most important thing to teach. I leave it up to you whether to mention this.

talk/basicconcepts/headersinterfaces.tex
```diff
@@ -73,3 +73,51 @@
...
+\end{frame}
+
+\begin{frame}[fragile]
+  \frametitlecpp[98]{Header / source separation}
+  \begin{block}{Implementation / source files}
+    \begin{itemize}
+      \item First include the header that is implemented
+      \item Include all headers that are needed to finish the implementation
+      \item Implement \emph{all} functions declared in the header
+    \end{itemize}
```
bernhardmgruber, line 108: Or just let clang-format sort them :)

talk/objectorientation/advancedoo.tex
```diff
@@ -1,5 +1,44 @@
 \subsection[advOO]{Advanced OO}
 
+\begin{frame}[fragile]
+  \frametitlecpp[98]{\texttt{this} keyword}
+  \begin{block}{How to know an object's address?}
+    \begin{itemize}
+    \item Sometimes we need the address of the current object
+    \item Or we need to pass our address / a reference to a different entity
+      (for example to implement operators, see later)
+    \item All class methods can use the keyword \cppinline{this}
```
bernhardmgruber, line 10: Nit: Only non-static ones.

talk/objectorientation/advancedoo.tex
```diff
@@ -1,5 +1,44 @@
...
+      \begin{itemize}
+        \item It returns the address of the current object
+        \item Its type is \cppinline{T*} in the methods of a class {\ttfamily T}
```
bernhardmgruber, line 13: For `const` member functions it would be `const T*`. But you can just mention that.

talk/objectorientation/advancedoo.tex
```diff
@@ -84,6 +123,53 @@
   \end{multicols}
 \end{frame}
 
+
+\begin{frame}[fragile]
+  \frametitlecpp[98]{Polymorphism and object slicing}
+  \begin{block}{What happens if we assign without reference or pointers?}
+    \begin{itemize}
+    \item The data members of the derived object are ``sliced off''
+    \item This is most of the time not the intention
```
bernhardmgruber, line 132: Maybe word more concisely
```suggestion
    \item This is usually wrong
```

talk/objectorientation/objectsclasses.tex
```diff
@@ -49,16 +52,46 @@
...
       }
     \end{tikzpicture}
+    \pause
+    \null \vfill
+    {
+      \setlength\textwidth\columnwidth
+    \begin{block}{struct vs.\ class}
+      \begin{itemize}
+        \item In structs, every member is accessible
+        \item In classes, members are ``private'' unless made public
```
bernhardmgruber, line 91:
```suggestion
        \item In structs, every member and base is ``public''
        \item In classes, they are ``private'' unless made public 
```

bernhardmgruber (commented):
Great work, here is some feedback:

talk/basicconcepts/headersinterfaces.tex
sponce: Remember we are in the basic part here. I would not enter this debate, especially that there is not even a consensus on the best practice here.

talk/basicconcepts/headersinterfaces.tex
sponce: `inline` keyword comes later, and only in the advanced course. Vectorization is not even mentioned in this course I believe, so I would skip this comment here.

sponce:
@hageboeck can you rebase so that pdf can be built and we can see the result ?

talk/objectorientation/advancedoo.tex
hageboeck: True, but maybe we are awake enough to say this. This woulda allow us to explain what the difference between static and and non-static methods is.

talk/objectorientation/advancedoo.tex
hageboeck: You wrote that line initially, so I'm happy that you are still almost happy with it. 🙂

talk/objectorientation/objectsclasses.tex
hageboeck: We are just before the slide where we deal with inheritance here. I put this box to not confuse people why we say class al they time, but the keyword is `struct`.

hageboeck:
> can you rebase so that pdf can be built and we can see the result ?

@sponce The CI completed now.

talk/objectorientation/objectsclasses.tex
bernhardmgruber: Alright, makes sense!

talk/objectorientation/advancedoo.tex
bernhardmgruber: :D

talk/basicconcepts/headersinterfaces.tex
bernhardmgruber: Sure, we would not discuss vectorization. I am just slighly afraid that the students would start to split *every* function between header and source and needlessly pessimize their code. But maybe I just have different preferences here. Also link-time-optimization (LTO) gets rid of the problem.

talk/objectorientation/objectsclasses.tex
```diff
@@ -49,16 +52,46 @@
...
+      };
+
+      MyFirstClass myObj;
+      myObj.a = 2; // error
+
+      // let's square a
+      myObj.squareA();
+    \end{cppcode*}
+    \end{overprint}
+      \columnbreak
```
sponce, line 75: This will come later, when it's useful. I do not see the added value to repeat it here. It will only take more time to present and blur the message

talk/objectorientation/objectsclasses.tex
```diff
@@ -49,16 +52,46 @@
...
+    \null \vfill
+    {
+      \setlength\textwidth\columnwidth
+    \begin{block}{struct vs.\ class}
+      \begin{itemize}
+        \item In structs, every member is accessible
+        \item In classes, members are ``private'' unless made public
+      \end{itemize}
+    \end{block}
+    }
```
sponce, line 94: Same

talk/objectorientation/objectsclasses.tex
```diff
@@ -109,48 +142,26 @@
...
-    \item For example to implement operators, see later
-    \item All class methods can use the keyword \cppinline{this}
+    \item Functions usually implemented in .cpp, outside of class, using the class name as ``namespace''
+    \item First include header being implemented, then everything that's needed for implementation
+    \item Keep includes and code in headers to a minimum (compiled many times)
       \begin{itemize}
-        \item It returns the address of the current object
-        \item Its type is \cppinline{T*} in the methods of a class {\ttfamily T}
+        \item Short member functions can be in the header
+        \item Some functions (templates, \cppinline{constexpr}) must be in the header
```
sponce, line 152: Previous version was less verbose and thus more readable. We should try to avoid plain sentences going on several lines and stick to one line bullets. I would also drop this change

talk/objectorientation/operators.tex
```diff
@@ -17,28 +17,36 @@
...
+      a-\textgreater & (a).operator-\textgreater() & cannot be non-member \\
+      a@ & (a).operator@(0) & operator@(a,0) \\
+      \hline
+    \end{tabular}
+
+    \small
+    \begin{cppcode*}{linenos=false}
+      possible operators: + - * / % ^ & | ~ ! = < >
+          += -= *= /= %= ^= &= |= << >> >>= <<=
+          == != <= >= <=> && || ++ -- , ->* -> () []
```
sponce, line 48: Very good idea to put the list here, no that much to have them all, but so that people realize how generic this is.
Out of curiosity, did you give them in the right order of precedence ?

sponce (commented):
I have the feeling some of the slides have reached the point where we start looping, and thus I have the feeling several changes are not worth it.

One point I'd like to underline is that we should have as few words as possible on the slides. And certainly no sentences spanning on several lines, or as few as possible. One line bullets are much more readable

talk/objectorientation/advancedoo.tex
```diff
@@ -84,6 +123,53 @@
...
+        \memorystruct{5}{6}{\tiny Drawable in f()}
+      }
+      \onslide<2>{
+        \memorypush{Shape.a}
+        \memorypush{...}
+        \memorystruct{5}{8}{\tiny Shape in catch block}
+      }
+    \end{tikzpicture}
+  \end{multicols}
+\end{frame}
```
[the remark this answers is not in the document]
hageboeck: @bernhardmgruber ?
You asked for discussing slicing. Did it go in the direction that you thought?

talk/objectorientation/operators.tex
hageboeck: I took the list from [cppreference](https://en.cppreference.com/w/cpp/language/operators). I don't know if it's in order, following our own advice not to try trusting into people knowing it. 🙂 
Looking up the preference I would say that they are not in order of preference.

talk/objectorientation/advancedoo.tex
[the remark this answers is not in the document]
bernhardmgruber: IIRC, a student asked about this during the course, or you explained it. I can't remember exactly.

I think we can remove the `try` and `catch` part of the example. Also:
```c++
      int f(Drawable d) {...}
      f(p);  //ok-ish
```
is not "ok-ish", because if Drawable is abstract, the code would not compile: You cannot create an instance of a Drawable.
Also, if Drawable has any virtual functions, they would no longer be called on the derived class, but on the base class. So you basically lose polymorphism and you are left with the functionality of a plain Drawable.

Maybe the picture of the memory layout is not needed. The main point is that students understand to not do it, and not how it works :)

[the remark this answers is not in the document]
hageboeck: OK, but looking at our discussion, this is clearly too advanced. I wasn't sure that we needed it, just added it because Bernhard wrote it down. Bernhard seems to not be sure why we wrote it down, so maybe the easiest is to really drop the slide.
If the students ask again, we can always pull up godbolt and do it on the fly.

OK to drop it?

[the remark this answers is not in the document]
bernhardmgruber: > Bernhard wrote it down. Bernhard seems to not be sure why we wrote it down

I wrote it down because the discussion came up during the course. One way or another. So I thought: let's add a slide for that!

> maybe the easiest is to really drop the slide.

Maybe. I would prefer to mention slicing somewhere, but I have no time to work on a slide :S So, up to you!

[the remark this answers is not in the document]
hageboeck: OK, removed now! 🙂

talk/objectorientation/objectsclasses.tex
hageboeck: I will remove this now.

hageboeck:
The parts where we couldn't agree are removed now. I think that should be enough now to close #428.

talk/basicconcepts/headersinterfaces.tex
```diff
@@ -73,3 +73,20 @@
...
+    \begin{itemize}
+      \item Headers should contain declarations of functions / classes
+      \item Might be included/compiled many times
+      \item Good to keep them short
+      \item Minimise \cppinline{#include} statements
+      \item Put long code in implementation files. Exceptions:
+        \begin{itemize}
+          \item Short functions
+          \item Templates and constexpr functions
+        \end{itemize}
```
sponce, line 89: Maybe add a line that headers are not mandatory and should not be created just for the sake of it. It's only for interfaces that are used somewhere else.

sponce (approved):

PR #509, opened 2024-01-19
Add an exercise on UBSan

### LINKED ISSUE #243: UBSan exercise
Use UBSan to run broken examples.

hageboeck:
Add a bug-ridden program that runs mostly without observable issues, but is full of undefined behaviour.

Using UBSan, all of these issues get revealed at run time.

Fix #243

### DIFF
--- a/exercises/ubsan/solution/undefinedBehaviour.cpp
+++ b/exercises/ubsan/solution/undefinedBehaviour.cpp
@@ -1,0 +1,60 @@
+#include <iostream>
+#include <limits>
+
+// You don't need to change any code in these structs:
+struct Base {
+  virtual void print() = 0;
+  virtual ~Base() {}
+};
+struct Derived1 : public Base {
+  void print() override {
+    std::cout << "Derived1::print()\n";
+  }
+};
+struct Derived2 : public Base {
+  void print() override {
+    std::cout << "Derived2::print()\n";
+  }
+};
+
+
+/**
+ * ***************
+ *  Instructions:
+ * ***************
+ *
+ * Compile and run this program using a compiler invocation like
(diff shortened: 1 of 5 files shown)


bernhardmgruber (approved):
I gave it a quick look and the example seems fine!

Great addition, thank you!

exercises/ubsan/solution/undefinedBehaviour.cpp
```diff
@@ -0,0 +1,60 @@
...
+ * Recompile with "-fsanitize=undefined", and observe that almost every second line contains a serious bug.
+ * Try to understand what's wrong.
+ */
+
+int runTests() {
+  int arr[] = {1, 2, 3, 4};
+  std::cout << "arr[4]=" << arr[4] << "\n";             // Array overrun
+
+  unsigned short s = 1;
+  std::cout << "s << 33=" << (s << 33) << "\n";         // We cannot shift a type with 32 bits by 33 bits.
```
sponce, line 43: And even less a 16 bits one !

sponce (approved):

exercises/ubsan/solution/undefinedBehaviour.cpp
hageboeck: Interestingly (@sponce), it didn't trigger when I shifted by 17 bits. So I read up on cppreference, and it turns out the integral promotion is applied before shifting. The right-hand side is `int`, so it was correct ... 🙂

sponce: indeed, that's interesting. But then I would use an unsigned int as I do not want to have to explain these details here.

hageboeck: OK, good suggestion. I merged in the mean, time, but I will quickly edit the file in-place now.

PR #514, opened 2024-03-25
Things from the 10th C++ Course

hageboeck:
Things noticed during the lecture:

1. There was some confusion regarding the for loop syntax: semicolon vs. comma. Here is an attempt to clarify, but I'm not sure if we are making it too complicated now.
2. ... TBC as we go on


sponce:
Tackled by #515

PR #517, opened 2024-03-28
Fix a crash in constplay exercise

hageboeck:
Although the exercise is mostly about battling with the compiler, once people finish this work, they will encounter a crash when running the program.
Here, we init the two pointers to &a and &b to create a runnable program.

@sponce Were the students supposed to fix the crash as well, or should we limit it to "just" compile it and then it should run without errors?


sponce:
I do not think this exercise was originally intended as something you can run. It was mostly to make the compiler happy.

PR #521, opened 2024-08-13
Make explicit statement that C++ compiled "to native machine code"

mosolovsa:
All main implementation of the languages in the example (Java, C# and Python) are actually compiled to bytecode. IMHO it's better to clearly add explicit statement that C++ is compiled "to native machine code" unlike this languages to avoid ambiguity.

### DIFF
--- a/talk/introduction/goals.tex
+++ b/talk/introduction/goals.tex
@@ -13,7 +13,7 @@
   \pause
   \begin{block}{Fast}
     \begin{itemize}
-    \item compiled (unlike Java, C\#, Python, ...)
+    \item compiled to native machine code (unlike Java, C\#, Python, ...)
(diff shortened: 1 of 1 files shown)


talk/introduction/goals.tex
```diff
@@ -13,7 +13,7 @@
   \pause
   \begin{block}{Fast}
     \begin{itemize}
-    \item compiled (unlike Java, C\#, Python, ...)
+    \item compiled to native machine code (unlike Java, C\#, Python, ...)
```
sponce, line 16: This change screws the layout of the slide (line is too long now). I've tested with
```suggestion
    \item compiled to native machine code
      \begin{itemize}
      \item unlike Java, C\#, Python, ...
      \end{itemize}
```
and this looks ok.

PR #523, opened 2024-08-21
Fix issue #352

### LINKED ISSUE #352: Things to fix on 1st/2nd/3rd day of advanced course
- Lambda part might be hasty
- Slide 243 (lambdas) goes to unique_ptr, custom deleters and cuda whereas smart pointers will be introduced much later.
- Template part is probably too complicated as well
- Didn't have enough time for RNGs
- Why are our spans not const in the "Usage" slide?

hageboeck:
In preparation for the next advanced course, I went through the list of suggestions in #352, and implemented most of them.

A few of these were not relevant enough on second thought, so I propose to close #352 after this PR.

Fix #352.

### DIFF
--- a/talk/tools/formatting.tex
+++ b/talk/tools/formatting.tex
@@ -17,7 +17,7 @@
 		\item \mintinline{bash}{clang-format -i <file.cpp>} (looks for .clang-format file)
 		\item \mintinline{bash}{git clang-format} (formats local changes)
 		\item \mintinline{bash}{git clang-format <ref>} (formats changes since git \textless{}ref\textgreater{})
-		\item Some editors/IDEs find a .clang-format file and adapt
+		\item Most editors/IDEs can find a .clang-format file and adapt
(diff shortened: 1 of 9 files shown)


talk/tools/formatting.tex
```diff
@@ -17,7 +17,7 @@
 		\item \mintinline{bash}{clang-format -i <file.cpp>} (looks for .clang-format file)
 		\item \mintinline{bash}{git clang-format} (formats local changes)
 		\item \mintinline{bash}{git clang-format <ref>} (formats changes since git \textless{}ref\textgreater{})
-		\item Some editors/IDEs find a .clang-format file and adapt
+		\item Some editors/IDEs (e.g. VSCode) find a .clang-format file and adapt
```
sponce, line 20:
```suggestion
		\item Most editors/IDEs can find a .clang-format file and adapt
```

sponce (approved):

PR #524, opened 2024-08-22
Added 2 slides on timing at the end of MoreSTL section

sponce:
Note that I believe we should make a Chapter from the  "MoreSTL" section. It starts to be too long and with many unrelated points

### DIFF
--- a/talk/morelanguage/morestl.tex
+++ b/talk/morelanguage/morestl.tex
@@ -424,3 +424,59 @@
     \end{cppcode*}
   \end{exampleblock}
 \end{frame}
+
+\begin{frame}[fragile]
+  \frametitle{\cpp date and time utilities \hfill \cpp11 / \cpp20}
+  \cppinline{std::chrono library} present in \cppinline{<chrono>} header
+  \begin{block}{Clocks}
+    \begin{itemize}
+    \item consists of a starting point (or epoch) and a tick rate
+      \begin{itemize}
+      \item E.g. January 1, 1970 and every second
+      \end{itemize}
+    \item C++ defines several clock type
+      \begin{itemize}
+      \item \href{https://en.cppreference.com/w/cpp/chrono/system_clock}{\color{blue!50!white} \cppinline{system_clock}} system time, aka wall clock time, or C time
+      \item \href{https://en.cppreference.com/w/cpp/chrono/steady_clock}{\color{blue!50!white} \cppinline{steady_clock}} monotonic but unrelated to wall clock time
+      \item \href{https://en.cppreference.com/w/cpp/chrono/high_resolution_clock}{\color{blue!50!white} \cppinline{high_resolution_clock}} clock with the smallest tick period
+      \end{itemize}
+    \end{itemize}
+  \end{block}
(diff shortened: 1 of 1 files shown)


talk/morelanguage/morestl.tex
```diff
@@ -424,3 +424,59 @@
...
+                    <std::chrono::milliseconds>(duration);
+      std::cout << "it took " << duration.count() << " ticks"
+                << ", that is " << millis.count() << " ms\n";
+    \end{cppcode*}
+  \end{exampleblockGB}
+  \pause
+  \begin{alertblock}{Warning}
+    \begin{itemize}
+    \item this does not measure the amount of CPU used !
+    \item neither the time spend on a CPU (think suspended threads)
```
hageboeck, line 479:
```suggestion
    \item neither the time spent on a CPU (think suspended threads)
```

talk/morelanguage/morestl.tex
```diff
@@ -424,3 +424,59 @@
...
+      std::chrono::high_resolution_clock clock;
+      clock::time_point start = clock::now();
+      ... // code to be timed
+      std::chrono::duration<float> ticks = clock::now() - start;
+
+      auto millis = std::chrono::duration_cast
+                    <std::chrono::milliseconds>(duration);
+      std::cout << "it took " << duration.count() << " ticks"
+                << ", that is " << millis.count() << " ms\n";
+    \end{cppcode*}
```
hageboeck, line 473: As an alternative, on could use `operator/(std::chrono::duration, ...)`. What do you think? The godbolt link shows an example of that.

```suggestion
  \begin{exampleblockGB}{How to measure the time spent in some code}{https://godbolt.org/z/avqj4hnMK}{\texttt{timing}}
    \small
    \begin{cppcode*}{gobble=2}
      #include <chrono>
      using namespace std::chrono_literals;

      std::chrono::high_resolution_clock clock;
      auto start = clock.now();
      std::this_thread::sleep_for(std::chrono::milliseconds(345));
      std::chrono::duration<float> ticks = clock.now() - start;

      std::cout << "it took " << ticks.count() << " ticks"
                << ", that is " << ticks / 1ms << " ms\n";
    \end{cppcode*}
```

hageboeck (approved):
Generally approved, but do you think it's worth replacing the duration cast by ` duration / 1ms` or `duration / std::chrono::milliseconds(1)` etc?

talk/morelanguage/morestl.tex
sponce: fixed

talk/morelanguage/morestl.tex
sponce: I hesitated, I did not have it initially, then added, then realized we did not present literals syntax and literal operators so it would look weird (did to me the first time), so I removed it again.

It's always the question of whether you present just recipes aka "black magic you should trust", or whether you build knowledge from scratch and explain everything you use before using it. I do not know

Btw : I've realized I forgot to push last version (with the fixes you had to do to compile). I've now pushed it.

hageboeck: OK, ready to merge! 👍

Now we can think about whether we want to present literals somewhere ... 🤷‍♂️ 🙂

sponce: :+1:  and #525 so that we do not forget

PR #538, opened 2024-10-02
Findings during the advanced course

hageboeck:


sponce (approved):

hageboeck:
> This seems to suffer from the missing minted in the OS image we use. I thought you did fix this @hageboeck. Or you've only tried ?

@sponce I tried a lot of things, but nothing worked. We can try again in a bit of time (when the texlive with the fix is in the container), or we can try to downgrade.

PR #539, opened 2024-10-02
Install extra texlive packages before building the slides

hageboeck:
Texlive packages seem to have disappeared from the standard image, so try installing them here.

**Edit**:
This actually goes back to a packaging error, see
https://github.com/gpoore/minted/issues/401

Will try next to reinstall using tlmgr.


hageboeck:
Looks like waiting fixed the problem.

PR #546, opened 2025-02-13
Used autogobble everywhere in minted by default

sponce:
And dropped (most of) the manual gobble= in the rest of the slides this has also fixed a couple of bad indentations here and there


hageboeck (approved):
Looks good!

A rebase should fix the build problem in the full course.

PR #549, opened 2025-02-18
Cleaned up .gitignore files and exercies compilation

sponce:
Eseentially cleaning up and splitting .gitignore

On the way, improved top level Makefile for compiling all tests in one go

Note that a few tests do not have cmake support. This should be fixed


hageboeck (approved):
LGTM!

PR #550, opened 2025-03-10
Findings from the essentials course

hageboeck:
Running list of improvements from the 12th instalment

### DIFF
--- a/exercises/basicTypes/basicTypes.cpp
+++ b/exercises/basicTypes/basicTypes.cpp
@@ -45,7 +45,7 @@ int main() {
   bool condition1 = false;
   bool condition2 = true;
   print( alwaysTrue && condition1 && condition2 );
-  print( alwaysTrue || condition1 && condition2 );  // Q: Why does operator precedence render this expression useless?
+  print( alwaysTrue || condition1 && condition2 );  // Q: Are this and the following expressions useful?
   print( alwaysTrue && condition1 || condition2 );
   print(condition1 != condition1);                  // Q: What is the difference between this and the following expression?
   print(condition2 = !condition2);
--- a/exercises/basicTypes/solution/basicTypes.sol.cpp
+++ b/exercises/basicTypes/solution/basicTypes.sol.cpp
@@ -45,8 +45,9 @@ int main() {
   bool condition1 = false;
   bool condition2 = true;
   print( alwaysTrue && condition1 && condition2 );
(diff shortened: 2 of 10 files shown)


chavid:
The final exercise (loospRefsAuto) was somehow redundant with the third one (functions).

chavid:
It would help to have more explicit recommendations on the required minimal version of g++ and clang++... which would require an easy way to check that a given version is ok for all exercises ? Perhaps already done in the CI.

chavid:
I see (only now) that there are many installation instructions in the exercices/README. Fine ! Is it somehow up to date ? Probably it deserves a dedicated file that we may review regularly.

exercises/basicTypes/basicTypes.cpp
chavid, line 1: Don´t you think that the lines 52-54 are an erroneous copy/paste of 47-49 ?

exercises/basicTypes/solution/basicTypes.sol.cpp
chavid, line 1: Don´t you think that the lines 52-54 are an erroneous copy/paste of 47-49 ?

hageboeck: Yes, this is fixed now.

hageboeck:
> Side comment: the print helper is handy but may be frightening for beginners.

I added a NOTE that it's not part of the exercise.

> It would help to have more explicit recommendations on the required minimal version of g++ and clang++... which would require an easy way to check that a given version is ok for all exercises ? Perhaps already done in the CI.

The CI already checks that exercises compile with the respective versions. You are right that we have to remember to update this part
https://github.com/hsf-training/cpluspluscourse/blob/cde47b7b45cbd7f81014d7e3511e444031737f1f/exercises/README.md?plain=1#L23-L28
when we move to C++23.

The `check_setup.sh` also checks compiler versions:
https://github.com/hsf-training/cpluspluscourse/blob/cde47b7b45cbd7f81014d7e3511e444031737f1f/exercises/check_setup.sh#L38-L39

exercises/basicTypes/basicTypes.cpp
chavid: Oh, I missed that ! I said that =! was an assignment, but did not realize one can see it with these three lines. Ok. Agreed,

sponce (approved):

PR #555, opened 2025-03-12
Adding a slide on chaining operators

sponce:


hageboeck (approved):
Thanks! Like this, we are also not violating the rule of 3, any more, because we didn't show the operator.

